PR #42, opened 2020-02-06
Add SimpleCov

kmahern:
#### What
Add simplecov

#### Ticket
https://dsdmoj.atlassian.net/browse/CACP-163

#### Why
To give details of test coverage


bin/rails
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+if ENV['RAILS_ENV'] == 'test'
+  require 'simplecov'
+  SimpleCov.start 'rails'
+  puts "required simplecov"
```
jsugarman: unneeded comment?

.gitignore
```diff
@@ -37,3 +37,6 @@ Brewfile.lock.json
 .env
 .env.local
 .env.**.local
+
+# ignore simplecov output
+/coverage
```
jsugarman: also add to `.dockerignore` 
[example](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/.dockerignore#L21)

spec/spec_helper.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'simplecov'
```
jsugarman: I would put this is rails_helper instead as best to keep spec_helper slim.

also consider adding config to exclude files from coverage 
[example config](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/spec/rails_helper.rb#L15-L42)

.gitignore
jsugarman: forget that, already in `.dockerignore`

spec/spec_helper.rb
kmahern: 👍

bin/rails
jsugarman: did not realise this from 
see https://github.com/colszowka/simplecov#getting-started

PR #43, opened 2020-02-06
Add govuk notify

kmahern:
#### What
Send devise emails from govuk notify.

#### Ticket

https://dsdmoj.atlassian.net/browse/CACP-122

#### Why
In order to notify users of password changes, email changes and password resets.

#### How

--------


app/mailers/devise_mailer.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class DeviseMailer < GovukNotifyRails::Mailer
+  include Devise::Controllers::UrlHelpers
+  include ActionView::Helpers::DateHelper
+
+  def reset_password_instructions(user, token, _opts = {})
+    set_template('7314475f-62b0-41bf-8aff-768427cba456')
```
jsugarman: extract template ids to method, constant or mixin, can pair on this:

```
  # example only
  GOVUK_NOTIFY_TEMPLATES = {
    reset_password_instructions: 'that-long-id',
    emailed_changed: 'that-long-id',
   }
```
and use
```
set_template(GOVUK_NOTIFY_TEMPLATES[:reset_password_instructions))

```

config/environments/production.rb
```diff
@@ -103,4 +103,9 @@
   # config.active_record.database_selector = { delay: 2.seconds }
   # config.active_record.database_resolver = ActiveRecord::Middleware::DatabaseSelector::Resolver
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
+
+  # Don't care if the mailer can't send.
+  config.action_mailer.raise_delivery_errors = false
```
jsugarman: we probably do want to raise errors?

app/mailers/devise_mailer.rb
kmahern: 👍

app/controllers/users_controller.rb
```diff
@@ -14,7 +16,7 @@ def create
     authorize!(:create, @user)
 
     if @user.save
-      @user.send_reset_password_instructions
+      deliver_reset_password_instructions(@user)
```
jsugarman: it would be good to have this method called form the user object

app/mailers/devise/mailer.rb
```diff
@@ -0,0 +1,52 @@
...
+      mail(to: user.email)
+    end
+
+    def email_changed(user, _opts = {})
+      set_template(GOVUK_NOTIFY_TEMPLATES[:email_changed])
+      set_personalisation(
+        user_full_name: user.name,
+        user_new_email: user.email
+      )
+      mail(to: user.email_before_last_save)
```
jsugarman: **Not a blocker** according to devise `email_was`, the default, should work 

https://github.com/heartcombo/devise/blob/a17abad57a5d97594701cee7eac072170f739313/lib/devise/models/database_authenticatable.rb#L176-L186

config/environments/test.rb
```diff
@@ -21,6 +21,9 @@
     'Cache-Control' => "public, max-age=#{1.hour.to_i}"
   }
 
+  # Don't care if the mailer can't send.
+  config.action_mailer.raise_delivery_errors = false
```
jsugarman: should match whatever we use in production

jsugarman:
could also add https://github.com/ministryofjustice/govuk_notify_rails#mailer-previewing

app/controllers/users_controller.rb
kmahern: I have removed password_helpers and this method is now called from the user object.

kmahern:
Backgrounding of email generation is in background-GovukNotify branch.

PR #45, opened 2020-02-07
Refactor search to use strategy pattern

jsugarman:
#### What
Apply strategy pattern to search along with adaptors/queryers

#### Ticket

[CACP-169](https://dsdmoj.atlassian.net/browse/CACP-169)
[CACP-117](https://dsdmoj.atlassian.net/browse/CACP-117)

#### Why
As the main purpose of the app and with different
kinds of searches needing to be applied in future
iterations this pattern should, hopefully, make
extensibility easier.

#### How
Introduce the idea of an adaptor (or "queryor", could rename)
that handles particular kinds of searches against particular
resources/endpoints of the API.

### DIFF
--- a/app/controllers/searches_controller.rb
+++ b/app/controllers/searches_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'court_data_adaptor'
+
 class SearchesController < ApplicationController
   before_action :set_search_args
 
@@ -9,7 +11,7 @@ class SearchesController < ApplicationController
   end
 
   def create
-    @search = Search.new(query: @query, filter: @filter)
+    @search = Search.new(adaptor: @adaptor)
     authorize! :create, @search
 
     @results = @search.execute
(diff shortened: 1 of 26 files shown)


app/models/search.rb
```diff
@@ -31,28 +27,44 @@ def self.filters
...
-      defendant_search
-    else
-      raise CourtDataAdaptor::Resource::NotFound, ''
-    end
+    adaptor.call
   end
+end
 
-  private
+# class ResourceQuery
```
jsugarman: remove

lib/court_data_adaptor.rb
```diff
@@ -2,7 +2,11 @@
 
 require_relative 'court_data_adaptor/resource/base'
 require_relative 'court_data_adaptor/resource/error'
-require_relative 'court_data_adaptor/prosecution_case'
+require_relative 'court_data_adaptor/resource/queryable'
```
jsugarman: add loop to require all

app/controllers/searches_controller.rb
```diff
@@ -9,7 +11,7 @@ def new
   end
 
   def create
-    @search = Search.new(query: @query, filter: @filter)
+    @search = Search.new(adaptor: @adaptor)
```
archferns: should this perhaps be a read_only attr? 
`(adaptor: adaptor)`

jsugarman: do you mean something like this

```
def create
  @search = Search.new(adaptor: adaptor)
  ...
end

...

def adaptor
   case @filter
    when 'defendant'
      CourtDataAdaptor::Query::Defendant.new(term)
    else
      CourtDataAdaptor::Query::ProsecutionCase.new(term)
    end
  end
...

```

archferns: This should do too, or else, you might want to make it an `attr_reader`, so as to avoid using an instance var.

lib/court_data_adaptor.rb
jsugarman: going to leave like this as makes debugging easier, for now

app/controllers/searches_controller.rb
jsugarman: have added that now

PR #50, opened 2020-02-11
Add codeclimate

kmahern:
#### What

Add simplecov and codeclimate

#### Ticket

https://dsdmoj.atlassian.net/browse/CACP-163

#### Why

To support code quality

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,7 +197,6 @@ jobs:
     executor: test-executor
     steps:
       - checkout
-      - setup_remote_docker
       - load-code-climate
       - build-base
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -178,6 +197,8 @@ jobs:
     executor: test-executor
     steps:
       - checkout
+      - setup_remote_docker
```
jsugarman: not needed for test executor as we don't build or push it:
https://circleci.com/docs/2.0/building-docker-images/

> once setup_remote_docker is called, a new remote environment is created, and your primary container is configured to use it. All docker-related commands are also executed in your primary container, but building/pushing images and running containers happens in the remote Docker Engine. (line 10)

PR #51, opened 2020-02-11
Prevent i18n translation error

jsugarman:
Prevent nil:Nilclass error and i18n::ArgumentError for nils.

Returned objects from API/adaptor does not have a DoB.

### DIFF
--- a/app/views/searches/_results.html.haml
+++ b/app/views/searches/_results.html.haml
@@ -17,4 +17,4 @@
           %td.govuk-table__cell= result.prosecution_case_reference
           %td.govuk-table__cell= result.first_name
           %td.govuk-table__cell= result.last_name
-          %td.govuk-table__cell= l(result.date_of_birth.to_date)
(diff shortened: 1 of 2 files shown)


app/views/searches/_results.html.haml
```diff
@@ -17,4 +17,4 @@
           %td.govuk-table__cell= result.prosecution_case_reference
           %td.govuk-table__cell= result.first_name
           %td.govuk-table__cell= result.last_name
-          %td.govuk-table__cell= l(result.date_of_birth.to_date)
+          %td.govuk-table__cell= l(result.date_of_birth.to_date) if result.date_of_birth
```
naseberry: This looks good, however, based on my knowledge of the data structure and search, shouldn't a DoB always be available?

jsugarman: should be, but code defensively. came up through use of local version of mock with factory generated values. i18n throws errors on the localize method with nil args

PR #78, opened 2020-03-04
Reconnect with fresh token on query, if required

jsugarman:
Fix unauthorized access on bearer token

#### What
This prevents unauthorized errors after token expiry (on master)

#### Ticket

[CACP-236](https://dsdmoj.atlassian.net/browse/CACP-236)

#### Why
adaptor implements a two hour expiry and the client is not currently retrieving a fresh token.
note that `client_credentials` do NOT implement `refresh_token` logic.

#### How
Reconnect with a fresh token on submission of a query. the retrieving of new token only occurs if the old one has expired - I believe Faraday::OAuth2 is doing this "magically"

### DIFF
--- a/lib/court_data_adaptor/resource/queryable.rb
+++ b/lib/court_data_adaptor/resource/queryable.rb
@@ -19,6 +19,13 @@ module CourtDataAdaptor
         def resource
           self.class.resource
         end
+
+        def refresh_token_if_required!
+          resource.connection(true) do |conn|
+            conn.use FaradayMiddleware::OAuth2, resource.client.bearer_token, token_type: :bearer
+            # conn.use Faraday::Response::Logger
+          end
+        end
       end
     end
   end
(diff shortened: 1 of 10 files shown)


lib/court_data_adaptor/resource/queryable.rb
```diff
@@ -19,6 +19,13 @@ module InstanceMethods
         def resource
           self.class.resource
         end
+
+        def refresh_token_if_required!
+          resource.connection(true) do |conn|
```
archferns: Does this get called regardless of the token being active?

jsugarman: Not sure exactly what magic is going on in either Faraday::Oauth2 or JsonApiClient::Connection but, from my code, it only gets a new token (calls 'oauth/token') if current token has `expired?`. 

From my manual local testing (with a 1 minute expiry time in the adaptor locally 😁 ) it only requests a new token on submission of a query if the previous has expired, even though this code always runs.  I am not sure of the impact of this code running on every query but network traffic wise it does not add any more round trips (except one extra when a new token is required, every 2 hours)

archferns: fair enough, thanks for clarifying

PR #80, opened 2020-03-05
Add search by NI and ASN no.

jsugarman:
#### What
Enable defendant searching by national insurance
number or arrest summons number

#### Ticket

[CACP-176](https://dsdmoj.atlassian.net/browse/CACP-176)

#### Why
To enable case workers to search for defendants by other
reference numbers - Part of designs.

#### How
Add an additional filter option which submits
queries to adaptor prosecution_case endpoint but
with different query filters  - to be based on
regex of the term/word/ref submitted by the
user to determine if an NI or ASN number.

#### TODO
- [X] NI search
- [X] ASN search
- [X] case; whitespace insensitive - NI term parse?!
- [X] add filter
- [X] submit query, process and display response
- [X] display ASN on results
- [ ] validate ASN format?
- [ ] present user with assumed reference kind


kmahern:
Looks good, just a question from me. It looks like we are validating nino to determine whether the search should be on nino or ASN (ie if it isn't a fully valid nino the search will be on ASN) but do we also want to include any validation as to whether the user has actually entered a valid ASN or nino, e.g. to check they haven't made a typo and entered an invalid/impossible nino/ASN rather than a nino/ASN that doesn't exist in CP? Don't know if this was covered somewhere in the user research?

app/models/search.rb
```diff
@@ -11,42 +11,45 @@ class Search
...
-    adaptor.call
+    query.call
   end
 
   private
 
-  def adaptor
-    send("#{filter}_adaptor")
+  def query
+    send("#{filter}_query")
```
archferns: since `filter` is being set via user input, would it be advisable to restrict the values it can take?

app/models/search.rb
```diff
@@ -11,42 +11,45 @@ class Search
...
-        id: :case_reference,
-        name: I18n.t('search_filter.radio_case_reference_label'),
-        description: I18n.t('search_filter.radio_case_reference_label_hint')
-      ),
-      SearchFilter.new(
-        id: :defendant,
-        name: I18n.t('search_filter.radio_defendant_label'),
-        description: I18n.t('search_filter.radio_defendant_label_hint')
-      )
+      SearchFilter.new(id: :case_reference,
```
jsugarman: move to method

app/models/search.rb
jsugarman: will look into, although the value is set by radio options in the view and there is no injection vector to be concerned with, i believe, since any other value will just simply raise a NoMethodError (which will filter up to a 500)

- but could validate against search filter collection ids.

jsugarman:
> Looks good, just a (non-blocking) query from me. It looks like we are validating nino to determine whether the search should be on nino or ASN (ie if it isn't a fully valid nino the search will be on ASN) but do we also want to include any validation as to whether the user has actually entered a valid ASN or nino, e.g. to check they haven't made a typo and entered an invalid/impossible nino/ASN rather than a nino/ASN that doesn't exist in CP? Don't know if this was covered somewhere in the user research?

Yep, we/I had settled on the `if nino; else` logic because we/I am not sure what the format of ASN is but it is relatively complex, and the mock is not setup with valid formats in any event - so should be a separate ticket to avoid delay on this one (or wait til we can query the HMCTS UAT environment with accurate ASN format data)

https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/360582/cjs-data-standards-catalogue-5.0.pdf

app/models/search.rb
jsugarman: have added a basic validation against the filter attribute now, cheers

PR #91, opened 2020-03-16
Implement unlinked case flow

jsugarman:
#### What
Add the page flows and designs for unlinked cases

#### Ticket

[CACP-251](https://dsdmoj.atlassian.net/browse/CACP-251)

#### Why
Implement designs to enable users
to retrieve details of a case and individual
defendant.

- [X] Add breadcrumbs for search filter, search, case details and defendant details
- [X] Apply [GDS design system breadcrumbs](https://design-system.service.gov.uk/components/breadcrumbs/)
- [ ] determine and display whether case is linked --> separate PR
- [ ] add form for linking --> separate PR


app/controllers/searches_controller.rb
```diff
@@ -5,6 +5,21 @@ class SearchesController < ApplicationController
 
   rescue_from JsonApiClient::Errors::ConnectionError, with: :connection_error
 
+  add_breadcrumb 'Search filters', :new_search_filter_path
+  add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
+  # add_breadcrumb \
```
kmahern: Remove commented out code?

lib/gds_design_system_breadcrumb_builder.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# Custom BreadcrumbsOnRails::Breadcrumbs::Builder class
```
kmahern: Remove comments?

app/controllers/searches_controller.rb
jsugarman: 👍

lib/gds_design_system_breadcrumb_builder.rb
jsugarman: these i put here deliberately as info. Arguably all classes should have high level descriptions to give context and example usage.

app/views/prosecution_cases/show.html.haml
```diff
@@ -0,0 +1,22 @@
...
+.govuk-heading-m
+  = t('search.result.defendant.heading')
+%table.govuk-table
+  %caption.govuk-table__caption.govuk-visually-hidden
+    = t('search.result.caption')
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.name')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.date_of_birth')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.maat_number')
```
naseberry: DRY table headers by creating an array within the relevant controller, and looping within the view.

Apply to all tables.

lib/gds_design_system_breadcrumb_builder.rb
```diff
@@ -0,0 +1,31 @@
...
+                compute_name(element)
+              else
+                @context.link_to_unless_current(
+                  compute_name(element),
+                  compute_path(element),
+                  element.options.merge(class: 'govuk-breadcrumbs__link')
+                )
+              end
+
+    @context.content_tag(:li, content, class: 'govuk-breadcrumbs__list-item')
```
naseberry: `aria-current="page"` required on the `li` of the current page

app/views/prosecution_cases/show.html.haml
jsugarman: At the moment I have avoided too much DRYing out as things may change. Putting presentation logic in the controllers is not great, although there is already some. I will look at sharing partials and or using view helpers. a `presenter`could also solve this problem.

lib/gds_design_system_breadcrumb_builder.rb
jsugarman: 👍

jsugarman: have added a commit to handle this @naseberry

naseberry: :+1:

PR #95, opened 2020-03-17
Stop checking development yarn integrity error

jsugarman:
#### What
Stop checking development yarn integrity error

#### Why
prevents rails running

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -52,9 +52,6 @@ development:
   <<: *default
   compile: true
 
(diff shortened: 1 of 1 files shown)


config/webpacker.yml
```diff
@@ -53,7 +53,7 @@ development:
   compile: true
 
   # Verifies that correct packages and versions are installed by inspecting package.json, yarn.lock, and node_modules
-  check_yarn_integrity: true
+  check_yarn_integrity: false
```
naseberry: You can remove this line, as it will fallback to the default setting (Line: 9)

PR #96, opened 2020-03-17
check yarn integrity

naseberry:
#### What
Disable `check_yarn_integrity` on development environment.

#### Why
`yarn check --integrity` is run on `rails c`, thereby causing the rails console to err.

#### How
Remove the environment specific `check_yarn_integrity` setting and fallback to webpacker default.


naseberry:
Duplicate

PR #98, opened 2020-03-18
Add unlinked case design flow

jsugarman:
#### What
Enable users to view case and defendant details with breadcrumb navigation

#### Ticket

[CACP-251](https://dsdmoj.atlassian.net/browse/CACP-251)

#### Why
Inline with designs - to enable users to drill down their search

#### TODO

 - [X] update views inline with designs
 - [X] add GDS breadcrumbs
 - [X] add form (controller/view)
 - [x] retrieve linked status for defendant

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,5 @@
 # DO NOT PUT SECRETS HERE
+# use with `source .env` in terminal to
+# prevent ruby 2.7 deprecation warnings.
+#
 export RUBYOPT='-W:no-deprecated'
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+# DO NOT PUT SECRETS HERE
+export RUBYOPT='-W:no-deprecated'
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ Brewfile.lock.json
 
 # dotenv (gem) ignores
 # check https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
-# .env
 .env.local
 .env.**.local
 
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,7 @@ Brewfile.lock.json
 
 # dotenv (gem) ignores
 # check https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
-.env
+# .env
 .env.local
 .env.**.local
 
--- a/app/controllers/defendants_controller.rb
+++ b/app/controllers/defendants_controller.rb
@@ -3,7 +3,7 @@
 class DefendantsController < ApplicationController
   before_action :load_and_authorize_search
 
-  add_breadcrumb 'Search filters', :new_search_filter_path
+  add_breadcrumb :new_search_filter_name, :new_search_filter_path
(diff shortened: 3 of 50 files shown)


.env
```diff
@@ -0,0 +1,2 @@
+# DO NOT PUT SECRETS HERE
+export RUBYOPT='-W:no-deprecated'
```
jsugarman: linebreak

.gitignore
```diff
@@ -34,7 +34,7 @@ Brewfile.lock.json
 
 # dotenv (gem) ignores
 # check https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
-.env
+# .env
```
jsugarman: remove

app/controllers/laa_references_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
```
jsugarman: remove or spec out and wireup

app/controllers/defendants_controller.rb
```diff
@@ -3,9 +3,9 @@
 class DefendantsController < ApplicationController
   before_action :load_and_authorize_search
 
-  add_breadcrumb 'Search filters', :new_search_filter_path
+  add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
   add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
-  add_breadcrumb 'Case details',
+  add_breadcrumb (proc { |v| v.prosecution_case_name(v.controller.defendant.prosecution_case_reference) }),
```
naseberry: Not part of the actual user journey, is this  required?

jsugarman: This adds the "Case ACASEREF" breadcrumb when on the defendants page so is part of the user journey. Are you referring to If the user skips and goes straight to defendant it also gets inserted - this is true but Dan liked that (and technically would require additional work to remove)

spec/fixtures/vcr_cassettes/spec/features/breadcrumbs_spec.yml
```diff
@@ -208,4 +208,55 @@ http_interactions:
...
+      string: ''
+    headers:
+      User-Agent:
+      - Faraday v0.17.3
+      Content-Type:
+      - application/vnd.api+json
+      Accept:
+      - application/vnd.api+json
+      Authorization:
+      - Bearer wXmu2gFJnWvJn6eIgDc-6WJChw68wJPxQz7ZVfaQD84
```
archferns: Although this is local, would be good to [filter sensitive data](https://relishapp.com/vcr/vcr/docs/configuration/filter-sensitive-data)

jsugarman: 👍 yeah, i noticed that earlier. you are right this is just local so does not matter but should remove for when we start recording against actual envs. will put a vcr filter in

jsugarman: @archferns have added filter and rerecorded all requests

PR #114, opened 2020-03-31
Add sentry

kmahern:
#### What
Add Sentry to track and log unhandled exceptions

#### Ticket
https://dsdmoj.atlassian.net/browse/CACP-263

#### Why
To track and log unhandled exceptions for live service

#### How
Created DSNs here https://sentry.service.dsd.io/mojds/ for Staging and Production environments
Added DSNs to secrets file
Added config initializer for Raven.

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


config/initializers/raven.rb
```diff
@@ -0,0 +1,11 @@
...
+
+require 'raven'
+
+if %w[production].include?(Rails.env) && ENV['SENTRY_DSN'].present?
+  Raven.configure do |config|
+    config.dsn = ENV['SENTRY_DSN']
+    config.environments = %w[production]
+    config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
+  end
+end
```
jsugarman: Our setup means all our server instances are production environments, so...
```suggestion
if Rails.env.eql?('production') && ENV['SENTRY_DSN'].present?
  Raven.configure do |config|
    config.dsn = ENV['SENTRY_DSN']
    config.environments = Rails.env
    config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
  end
end
```

kmahern: 👍

PR #121, opened 2020-04-03
Link defendant - associate a MAAT reference with a defendant

jsugarman:
#### What
Enable linking of defendant with a MAAT reference

#### Ticket

[CACP-125](https://dsdmoj.atlassian.net/browse/CACP-125)

#### Why
Required by users to process claims for remuneration

#### How
Provide form submission that calls relevant adaptor endpoint for creation
of a "link" between a defendant and a MAAT reference.


app/views/defendants/_representation_orders.html.haml
```diff
@@ -0,0 +1,19 @@
...
+  = t('search.result.representation_order.heading')
+
+%table.govuk-table
+  %caption.govuk-table__caption.govuk-visually-hidden
+    = t('search.result.representation_order.caption')
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.representation_order.granted_at')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.representation_order.maat_number')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.representation_order.provider')
```
mpw5, line 11: More of a wider design question, not something to be addressed now, but is there any plan to display the status of a representation order here? A rep order can have a status of granted, refused, or withdrawn (plus variations on those). Do we want to add an extra column to display that information to the caseworker? At the moment the use of 'Granted at' as a column header would suggest to a casworker that any rep order they see there has been granted, but that might not be the case.

jsugarman: good point. can you raise with Dan

PR #163, opened 2020-06-03
Unlink a case from MAAT 

kjdchapman:
We can mark a particular case as "linked" to a MAAT ID, so that it'll get more information from HMCTS' Common Platform.

Sometimes, we'll do this accidentally, or get the wrong ID (since it's free-typed) - this change allows users to reverse the process.

The ["Unlink a case"](https://dsdmoj.atlassian.net/browse/CACP-128) ticket on JIRA.

Questions I have:
- Is it good to merge this and worry about getting the right data across in a follow-up?
- Does [removing this relation](https://github.com/ministryofjustice/laa-court-data-ui/pull/163/files#diff-0c312595236dd3d6c044a0a6a5c6a832L6) have any unintended side effects?

### DIFF
--- a/app/controllers/defendants_controller.rb
+++ b/app/controllers/defendants_controller.rb
@@ -13,6 +13,16 @@ class DefendantsController < ApplicationController
                    defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
   end
 
+  def remove_link
+    defendant.update(
+      user_name: 'example',
+      unlink_reason_code: '',
+      unlink_reason_text: 'unknown'
+    )
+
+    redirect_to defendant_path(defendant.arrest_summons_number)
+  end
+
   def defendant
     @defendant ||= @search.execute.first
   end
--- a/app/controllers/defendants_controller.rb
+++ b/app/controllers/defendants_controller.rb
@@ -16,8 +16,8 @@ class DefendantsController < ApplicationController
   def remove_link
     defendant.update(
       user_name: 'example',
-      unlink_reason_code: '',
-      unlink_reason_text: 'unknown'
+      unlink_reason_code: 1,
+      unlink_reason_text: 'Wrong MAAT ID'
     )
 
     redirect_to defendant_path(defendant.arrest_summons_number)
--- a/lib/court_data_adaptor/resource/defendant.rb
+++ b/lib/court_data_adaptor/resource/defendant.rb
@@ -3,9 +3,11 @@
 module CourtDataAdaptor
   module Resource
     class Defendant < Base
-      belongs_to :prosecution_case
       property :prosecution_case_reference, type: :string
       property :name, type: :string
+      property :user_name
+      property :unlink_reason_code
+      property :unlink_reason_text
 
       def linked?
         maat_reference&.present?
--- a/app/views/laa_references/_unlink.html.haml
+++ b/app/views/laa_references/_unlink.html.haml
@@ -1,4 +1,4 @@
-= link_to t('laa_reference.unlink.label'), '', class: 'govuk-link'
+= link_to t('laa_reference.unlink.label'), remove_link_defendant_path(@defendant.arrest_summons_number), method: :delete, class: 'govuk-link'
 %br
 %br
 .govuk-warning-text
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,8 +22,12 @@ Rails.application.routes.draw do
   end
 
   resources :prosecution_cases, only: %i[show]
-  resources :defendants, only: %i[show]
-  resources :laa_references, only: %i[create destroy]
+
+  resources :defendants, only: %i[show] do
+    delete :remove_link, on: :member
+  end
+
+  resources :laa_references, only: %i[create]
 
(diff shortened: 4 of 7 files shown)


app/controllers/defendants_controller.rb
```diff
@@ -13,6 +13,16 @@ def show
                    defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
   end
 
+  def remove_link
+    defendant.update(
+      user_name: 'example',
```
jsugarman, line 18: If is this is intended to be the user that unlinked, can we not just send `current_user.email` (what is the expected format of user_name (i would push for email address if possible but think there may be issues with that related to MAATs username logic)

app/controllers/defendants_controller.rb
```diff
@@ -13,6 +13,16 @@ def show
...
+      unlink_reason_code: 1,
+      unlink_reason_text: 'Wrong MAAT ID'
```
jsugarman, line 20: regarding using hard-coded values and your question in the PR i would say this if fine for now. 

are their tickets for:
1. retrieving a list of reason codes/descriptions (possibly on the fly from HMCTS via the adaptor)
2. adding fields to the form for user to select  a reason

app/controllers/defendants_controller.rb
```diff
@@ -13,6 +13,16 @@ def show
                    defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
   end
 
+  def remove_link
```
jsugarman, line 16: i think my original hope for unlinking (it all comes back now), was to use the [laa_references_controller](https://github.com/ministryofjustice/laa-court-data-ui/blob/master/app/controllers/laa_references_controller.rb)  create and destroy actions to enable a Restful approach. @archferns would this be something you think is worth doing?

However, looking at the current state of the adaptor's API for [laa_references](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/controllers/api/internal/v1/laa_references_controller.rb) and [defendants](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/controllers/api/internal/v1/defendants_controller.rb#L7) it looks like this is a sensible alternative. 

Would it be clearer, however, to have the link call an update action in the controller?

very roughly and assuming routes changed
```
link_to "Unlink whatever", defendant_path(@defendant, current_user), method: :put, class: 'govuk-link'
```

lib/court_data_adaptor/resource/defendant.rb
```diff
@@ -3,9 +3,11 @@
 module CourtDataAdaptor
   module Resource
     class Defendant < Base
-      belongs_to :prosecution_case
       property :prosecution_case_reference, type: :string
       property :name, type: :string
+      property :user_name
```
jsugarman, line 8: as discussed and to answer 
> Does removing this relation have any unintended side effects?

I do not believe so 👍

app/controllers/defendants_controller.rb
kjdchapman: From looking at the adapter, I think [user names are limited to ten characters](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/contracts/unlink_defendant_contract.rb#L17-L19) - I did have this as email originally.

I wasn't sure what usernames are most useful, so wanted to skip that discussion, get this merged, and worry about it in a follow-up! Is your preference to create tickets for things like that?

app/controllers/defendants_controller.rb
kjdchapman: We've got https://dsdmoj.atlassian.net/browse/CBO-1270, though this doesn't cover the usernames part. 

I could add "Send the right usernames with unlink requests" as an extra ticket or add it to that, not sure what is best - I've asked Deepika/Anne-Marie on slack.

app/controllers/defendants_controller.rb
kjdchapman: I started off with this exact approach because it was the least surprising - especially as we create references with that controller! But as you note, the adaptor API doesn't work the way I expected - and having the `laa_references_controller` try and do something with a `defendant` felt less clear than this.

It may be worth feeding back to the adaptor team that this was a surprising thing for us? I'm worried about making them make changes when there's a looming deadline, though.

I thought using a non-RESTful route might help highlight the weirdness - but I'm not sure what is best here. I'll happily change it to `#update` if you think that it doesn't hide the surprise too much!

app/controllers/defendants_controller.rb
archferns: To clarify: the reason codes + username is not required by HMCTS, but only by our internal systems, as CP does not have the ability to unlink yet.

app/controllers/defendants_controller.rb
jsugarman: sounds good 👍

app/controllers/defendants_controller.rb
jsugarman: lets have a chat with the adaptor team, tomorrow if possible, and see if its a big ask. otherwise happy to go with this approach

PR #175, opened 2020-06-10
Unlink with unique username

maysakanoni:
Previously a generic username was sent as part of the unlinking. However this is not sufficient as it doesn't allow identification of the user who unlinked the case. There is a 10 char limit enforced by MLRA for the username which does not allow for using the user email as the username. Therefore,  a uniquely user reference will be used instead and VCD will be the source of truth for identifying the user. 

A primary key cannot be used to do this as there are potential security implications with using auto incrementing ids in this context. There are also concerns regarding the fact that primary ids are susceptible to change (so when an external service only has awareness of this id when communing about the user, then once this primary id changes in the db it can no longer identify the user ). Therefore, a unique user reference is used (which is not susceptible to change as primary id may be) to serve this need

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,10 +14,13 @@ class User < ApplicationRecord
           :validatable,
           :lockable
 
+  before_validation :generate_uuid
+
   accepts_roles :caseworker, :manager, :admin
 
   validates :email, confirmation: true
   validates :email_confirmation, presence: true, if: :email_changed?
+  validates :unique_user_reference, uniqueness: true
 
   def name
     "#{first_name} #{last_name}"
@@ -25,5 +28,13 @@ class User < ApplicationRecord
 
   def send_devise_notification(notification, *args)
     devise_mailer.send(notification, self, *args).deliver_later
+  end
+
+  def generate_uuid
+    return if unique_user_reference.present?
+    while unique_user_reference.blank?
+      ref = SecureRandom.hex(5)
+      self.unique_user_reference = ref if User.where(unique_user_reference: ref).blank?
+    end
   end
 end
--- a/db/migrate/20200611102258_add_unique_reference_to_user.rb
+++ b/db/migrate/20200611102258_add_unique_reference_to_user.rb
@@ -1,0 +1,6 @@
+class AddUniqueReferenceToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :unique_user_reference, :string, null: false
+    add_index :users, :unique_user_reference, unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_21_130904) do
+ActiveRecord::Schema.define(version: 2020_06_11_102258) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -34,8 +34,10 @@ ActiveRecord::Schema.define(version: 2020_01_21_130904) do
     t.string "first_name", null: false
     t.string "last_name", null: false
     t.string "roles", default: ["caseworker"], array: true
+    t.string "unique_user_reference", null: false
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
+    t.index ["unique_user_reference"], name: "index_users_on_unique_user_reference", unique: true
     t.index ["unlock_token"], name: "index_users_on_unlock_token", unique: true
   end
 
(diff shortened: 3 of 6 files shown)


app/models/user.rb
```diff
@@ -26,4 +29,8 @@ def name
   def send_devise_notification(notification, *args)
     devise_mailer.send(notification, self, *args).deliver_later
   end
+
+  def generate_uuid
+    self.unique_user_reference = SecureRandom.hex(5) if unique_user_reference.blank?
```
richrace, line 34: You'd need to add some logic here for checking if the reference has already been taken.

E.g. 

```ruby
  return if unique_user_reference.present?
  while unique_user_reference.blank?
    ref = SecureRandom.hex(5) 
    self.unique_user_reference = ref if User.where(unique_user_reference: ref).blank?
  end 
```

app/models/user.rb
```diff
@@ -14,10 +14,13 @@ class User < ApplicationRecord
           :validatable,
           :lockable
 
+  before_create :generate_uuid
```
richrace, line 17:
```suggestion
  before_validation :generate_uuid
```

db/migrate/20200611102258_add_unique_reference_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueReferenceToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :unique_user_reference, :string, null: false, unique: true
```
richrace, line 3: The schema file doesn't have the unique index. Might have to `add_index` separately.

app/models/user.rb
maysakanoni: this is what I was after, thank you!

db/migrate/20200611102258_add_unique_reference_to_user.rb
maysakanoni: oh yeah, didn't spot that

jsugarman:
Hi, Have not been part of conversations so not sure what has been discussed in any detail but my thoughts are:

1. I am a bit concerned about the use of VCD users being the source of truth for user identification. Has this been agreed in discussions with adaptor and MLRA lot. would it be a lot of work to amend MLRA to either have an email field (associated with a user) that can be added and use email instead to identify the MLRA user
3. if uuids are to be used and be sourced from VCD then the following could be used

 - `SecureRandom.uuid` is available instead of `SecureRandom.hex(5)` but not sure this is 
      needed in either case, see below
 -   `add_column :users, :unique_user_reference, :uuid, default: 'uuid_generate_v4()', index: true`
 -    could add second migration to enforce `null: false` (if that is not possible currently with the default option)

will add ^^ as comments to the PR for simplicity

app/models/user.rb
```diff
@@ -14,10 +14,13 @@ class User < ApplicationRecord
           :validatable,
           :lockable
 
+  before_validation :generate_uuid
```
jsugarman, line 17: have generally handled uuid generation on the database directly rather than the model

db/migrate/20200611102258_add_unique_reference_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueReferenceToUser < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :unique_user_reference, :string, null: false
+    add_index :users, :unique_user_reference, unique: true
```
jsugarman, line 4:
```suggestion
   add_column :users, :unique_user_reference, :uuid, default: 'uuid_generate_v4()', index: true
```

i'd be tempted to just call the column/field `uuid` as well
```suggestion
   add_column :users, :uuid, :uuid, default: 'uuid_generate_v4()', index: true
```

jsugarman: will need to enable a postgres extension if this is not already enabled

https://www.postgresql.org/docs/11/uuid-ossp.html

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_21_130904) do
+ActiveRecord::Schema.define(version: 2020_06_11_102258) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
```
jsugarman, line 16: To enable uuid generation on the DB you would need to add the postgres extension
```suggestion
  enable_extension "plpgsql"
  enable_extension "uuid-ossp"
```

jsugarman:
just realised you need to keep to 10 chars because of MLRA limitations. sorry

app/models/user.rb
jsugarman: currently this is not a UUID. i'd name it `generate_unique_ref` or somesuch

jsugarman:
closing in favour of #180 following discussion with devs and PMs

PR #194, opened 2020-06-24
Move sidekiq to worker pods

jsugarman:
#### What
create separate pods for workers

#### Ticket

N/A

#### Why

Offloads work to dedicated instances/pods
and works around the sidekiq 6 need for
a supervisor for running sidekiq daemonized,
as we currently do, which will enabled us
to upgrade sidekiq.

Also means we have dedicated sidekiq logs
to look at.

### DIFF
--- a/.k8s/dev/deployment-worker.yaml
+++ b/.k8s/dev/deployment-worker.yaml
@@ -1,0 +1,82 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: laa-court-data-ui-worker
+  namespace: laa-court-data-ui-dev
+spec:
+  replicas: 1
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 100%
+  selector:
+    matchLabels:
+      app: laa-court-data-ui
+      tier: worker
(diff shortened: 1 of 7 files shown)


lib/tasks/worker.rake
```diff
@@ -0,0 +1,7 @@
+namespace :worker do
+  desc 'Run all tests in application-defined order, intended for CI test suite run'
```
jsugarman, line 2: change description

lib/tasks/worker.rake
```diff
@@ -0,0 +1,7 @@
...
+  task :healthcheck => :environment do
+    raise StandardError unless `pgrep -f sidekiq`.present?
+  end
+end
+
```
jsugarman, line 7: kill extra space

.k8s/dev/deployment-worker.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: laa-court-data-ui-worker
+  namespace: laa-court-data-ui-dev
+spec:
+  replicas: 2
```
jsugarman, line 7: change to 1 replica

PR #208, opened 2020-07-07
[security] bump sidekiq 5.2.7 --> 6.10

jsugarman:
#### What

Bump sidekiq to version 6+

#### Why
Although not directly related to security,
using sidekiq 5 means we cannot bump rack
to (< 2.1) and rack has a vulnerability
that needs handling:

https://github.com/ministryofjustice/laa-court-data-ui/network/alert/Gemfile.lock/rack/open

#### Note 
Since we recently offloaded sidekiq server to worker
pods the previous issue with sidekiq upgrade (proper
daemonization) should no longer be an issue

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'puma', '~> 4.3'
 gem 'rails', '~> 6.0.3'
 gem 'redis', '~> 4.2.1'
 gem 'sentry-raven'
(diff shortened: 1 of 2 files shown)


jsugarman:
Have test this on dev and worker is still working fine (i.e. sending emails)

PR #216, opened 2020-07-09
Cbo 1358 apply error handling designs for linking v3

kmahern:
#### What
Add link_attempt model and associated validations to laa_references_controller.rb

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1358

#### Why
So user is clear what  they must do to correctly link a defendant on a case with a MAAT reference - and by following GDS standards for error presentation

#### How
Added link_attempt.rb model and associated validation errors. Because of the way that the linking and unlinking forms were previously nested on the same page, this also involved separating the pages displaying the two forms and tweaking routing so that a successful unlink redirects to new_laa_references_path, and a sucessful link to edit_defendants_path, as well as fixing the specs that this broke.


app/controllers/defendants_controller.rb
```diff
@@ -85,4 +73,8 @@ def set_unlink_attempt
                         UnlinkAttempt.new
                       end
   end
+
+  def defendant_identifier
+    @defendant.arrest_summons_number || @defendant.national_insurance_number
```
jsugarman, line 78: can this be `defendant.arrest_summons_number || defendant.national_insurance_number` and could it be moved to DefendantHelper concern?

app/controllers/laa_references_controller.rb
```diff
@@ -1,14 +1,36 @@
...
+      if link_laa_reference
+        flash[:notice] = I18n.t('laa_reference.link.success')
+        redirect_to edit_defendant_path(@defendant_identifier)
+      else
+        flash[:alert] = I18n.t('laa_reference.link.failure', error_messages: error_messages)
+        redirect_to new_laa_reference_path(id: @defendant_identifier)
+      end
     else
-      flash[:alert] = I18n.t('laa_reference.link.failure', error_messages: error_messages)
+      render :new
```
jsugarman, line 32: here you use symbol, other controller uses string. i'd choose one and apply to both, for consistency

app/controllers/laa_references_controller.rb
```diff
@@ -26,16 +48,26 @@ def laa_reference_params
     params.permit(
       :id,
       :defendant_id,
-      :maat_reference
+      link_attempt: %i[id defendant_id maat_reference defendant_identifier]
```
jsugarman, line 51: still a bit confusing why we need `:defendant_id` twice and then alsow have a `:defendant_identifier` - could rename defendant_identifier to exactly what it is - `defendant_asn_or_nino` - we are going to need to remove this part at some point innay events as CP does not guaruantee either ASN or NINO will exist, will all have to be id

app/models/link_attempt.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class LinkAttempt
+  include ActiveModel::Model
+  attr_accessor :id, :maat_reference, :defendant_id, :no_maat_id, :defendant_identifier
```
jsugarman, line 5: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/216/files#r454209317)

app/views/laa_references/new.html.haml
```diff
@@ -0,0 +1,3 @@
+= govuk_page_title(@defendant.name, t('generic.defendant'))
+
+= render('laa_references/form', locals: @defendant)
```
jsugarman, line 3: linebreak

app/views/prosecution_cases/show.html.haml
```diff
@@ -15,7 +15,10 @@
       - cse.defendants.each do |defendant|
         %tr.govuk-table__row
           %td.govuk-table__cell
-            = link_to defendant.name, edit_defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
+            - if defendant.linked?
```
jsugarman, line 18: consider using a view helper/decorator of some kind

app/views/results/_defendant.html.haml
```diff
@@ -12,7 +12,10 @@
     - @results.each do |defendant|
       %tr.govuk-table__row
         %td.govuk-table__cell
-          = link_to defendant.name, edit_defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
+          - if defendant.linked?
```
jsugarman, line 15: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/216/files#r454212957)

app/views/results/_prosecution_case.html.haml
```diff
@@ -13,7 +13,10 @@
       - cse.defendants.each do |defendant|
         %tr.govuk-table__row
           %td.govuk-table__cell
-            = link_to defendant.name, edit_defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
+            - if defendant.linked?
```
jsugarman, line 16: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/216/files#r454212957)

spec/requests/linking/link_defendant_maat_reference_spec.rb
```diff
@@ -6,11 +6,16 @@
...
-      id: nino,
-      defendant_id: defendant_id,
-      maat_reference: maat_reference
+      link_attempt:
+      {
+        id: nino,
+        defendant_id: defendant_id,
+        maat_reference: maat_reference,
+        defendant_identifier: defendant_identifier
+      }
```
jsugarman, line 18: if the defendant :id is the nino (or asn), why can we not just use that in the controllers generally. 

i'd rename the defendant_identifier, as previously described as this can help show up these oddities and/or duplications

app/controllers/defendants_controller.rb
kmahern: 👍 changed to defendant.arrest_summons_number || defendant.national_insurance_number and could it be moved to DefendantHelper concern.

app/controllers/laa_references_controller.rb
kmahern: 👍 I have now used string in both controllers

app/views/laa_references/new.html.haml
kmahern: 👍

app/views/prosecution_cases/show.html.haml
kmahern: 👍 created defendant_link_helper.rb

app/views/results/_defendant.html.haml
kmahern: 👍 created defendant_link_helper.rb

app/views/results/_prosecution_case.html.haml
kmahern: 👍 created defendant_link_helper.rb

spec/requests/linking/link_defendant_maat_reference_spec.rb
kmahern: 👍 I have removed defendant identifier and added defendant_asn_or_nino instead of id.

app/controllers/laa_references_controller.rb
kmahern: 👍 I have remove defendant_id from laa_reference_params and renamed defendant_identifier to defendant_asn_or_nino. I have also removed `f.hidden_field(:id), params[:id] ` from the form and replaced with `f.hidden_field(:defendant_asn_or_nino, value: (params[:id] || @defendant_asn_or_nino))`

app/controllers/concerns/defendant_helpers.rb
```diff
@@ -9,6 +9,10 @@ def defendant
 
   private
 
+  def defendant_identifier
```
jsugarman, line 12: `defendant_asn_or_nino`

app/controllers/concerns/defendant_helpers.rb
```diff
@@ -19,7 +23,7 @@ def set_defendant_if_required
   end
 
   def term
-    @defendant_identifier || defendant_params[:id]
+    @defendant_asn_or_nino || defendant_params[:id]
```
jsugarman, line 26:
```suggestion
    defendant_asn_or_nino || defendant_params[:id]
```

app/controllers/laa_references_controller.rb
```diff
@@ -23,13 +23,13 @@ def create
     if @link_attempt.valid?
       if link_laa_reference
         flash[:notice] = I18n.t('laa_reference.link.success')
-        redirect_to edit_defendant_path(@defendant_identifier)
+        redirect_to edit_defendant_path(@defendant_asn_or_nino)
```
jsugarman, line 26:
```suggestion
        redirect_to edit_defendant_path(defendant_asn_or_nino)
```

app/controllers/laa_references_controller.rb
```diff
@@ -23,13 +23,13 @@ def create
...
       else
         flash[:alert] = I18n.t('laa_reference.link.failure', error_messages: error_messages)
-        redirect_to new_laa_reference_path(id: @defendant_identifier)
+        redirect_to new_laa_reference_path(id: @defendant_asn_or_nino)
```
jsugarman, line 29:
```suggestion
        redirect_to new_laa_reference_path(id: defendant_asn_or_nino)
```

app/controllers/laa_references_controller.rb
```diff
@@ -58,8 +55,8 @@ def link_attempt_params
     laa_reference_params[:link_attempt].merge(no_maat_id: no_maat_id?)
   end
 
-  def defendant_identifier
-    @defendant_identifier = laa_reference_params[:id] || link_attempt_params[:defendant_identifier]
+  def defendant_asn_or_nino
+    @defendant_asn_or_nino = laa_reference_params[:id] || link_attempt_params[:defendant_asn_or_nino]
```
jsugarman, line 59: Does it break anything to use `||=` rather than '=' ? I am not sure of impact however

app/helpers/defendant_link_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module DefendantLinkHelper
+  def defendant_link_path(defendant)
+    if defendant.linked?
```
jsugarman, line 5: 🥇

jsugarman: You should unit test this though `spec/helpers/defendant_link_helper_spec.rb`

app/views/prosecution_cases/show.html.haml
```diff
@@ -15,10 +15,7 @@
       - cse.defendants.each do |defendant|
         %tr.govuk-table__row
           %td.govuk-table__cell
-            - if defendant.linked?
-              = link_to defendant.name, edit_defendant_path(defendant.arrest_summons_number || defendant.national_insurance_number)
-            - else
-              = link_to defendant.name, new_laa_reference_path(id: defendant.arrest_summons_number || defendant.national_insurance_number)
+            = link_to defendant.name, defendant_link_path(defendant)
```
jsugarman, line 18: 👍

app/helpers/defendant_link_helper.rb
jsugarman: I'd also rename this module to a more generic `DefendantHelper` and, to avoid name clashes and aid readability rename the DefendantHelper controller concern to something else

app/controllers/concerns/defendant_helpers.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module DefendantHelpers
```
jsugarman, line 3: I'd rename this since you now have two DefendantHelper(s) modules - DefendantSearchable perhaps.

TBH i don't think we should be using a concern - a service object would be better as it can be tested easily in isolation and is not so tightly coupled.

app/controllers/concerns/defendant_helpers.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def set_defendant_if_required
+    defendant
+  end
+
+  def term
+    @defendant_asn_or_nino || defendant_params[:id]
+  end
+
+  def add_defendant_case_breadcrumb
```
jsugarman, line 29: I find this method in here to be a little odd - maybe we should just repeat this call in the relevant controllers and remove from this concern. The concern should be for one thing and one thing only which to me would be enabling search of a defendant as [mentioned above](https://github.com/ministryofjustice/laa-court-data-ui/pull/216/files#r454958421)

app/controllers/defendants_controller.rb
```diff
@@ -1,30 +1,31 @@
 # frozen_string_literal: true
 
 class DefendantsController < ApplicationController
+  include DefendantHelpers
```
jsugarman, line 4: rename DefendantSearchable, as [mentioned above](https://github.com/ministryofjustice/laa-court-data-ui/pull/216/files#r454958421)

kmahern:
test

app/helpers/defendant_link_helper.rb
kmahern: 👍 I have created spec/helpers/defendant_helper_spec.rb and renamed defendant_link_helper.rb defendant_helper.rb

app/controllers/laa_references_controller.rb
kmahern: 👍 now using ||=

app/controllers/concerns/defendant_helpers.rb
kmahern: I cannot make this change, as it breaks linking. I think the issue is the defendant_controller and laa_references_controller work slightly differently. In the laa_references_controller I have had to introduce @defendant_asn_or_nino as params[:id] is sometimes null, for example after failed validation, hence `f.hidden_field(:defendant_asn_or_nino, value: (params[:id] || @defendant_asn_or_nino))` in the view. In laa_references_controller I need to use @defendant_asn_or_nino in load_and_authorize_search to set the defendant (in order to then get the id to send to the adaptor for linking etc.)

app/controllers/defendants_controller.rb
kmahern: 👍 Renamed DefendantSearchable

app/controllers/concerns/defendant_helpers.rb
kmahern: I have renamed defendant_helper to defendant_searchable 👍 

Regarding your comment about using service object rather than concern, I take your point about not being able to test the helper in isolation, however, I introduced the helper in order to avoid duplicating some shared code and I am not sure in my own mind how a service object would work in this instance. Would you be happy to have a separate spike ticket to investigate using a service object perhaps this could also include the work to remove dependencies on defendant asn and nino?

app/controllers/concerns/defendant_helpers.rb
kmahern: see comment below vvv

app/controllers/laa_references_controller.rb
kmahern: see comment above ^^^

app/controllers/laa_references_controller.rb
kmahern: see comment above ^^^

app/controllers/concerns/defendant_helpers.rb
kmahern: I have moved the add_defendant_case_breadcrumb back into the two controllers but this is causing code climate to fail - should we just ignore this or attempt to refactor in a different way?

app/controllers/concerns/defendant_helpers.rb
jsugarman: what is value of id? is it `uuid` or `asn_or_nino`?

app/controllers/concerns/defendant_helpers.rb
jsugarman: yes, we should do this. can you add it as a checklist item on the existing ticket for stopping reliance or asn and nino, and link that to a separate ticket to break it into a service object (as we may or may not get time to do that)

app/controllers/concerns/defendant_helpers.rb
jsugarman: ignore for now i'd say

app/controllers/concerns/defendant_searchable.rb
```diff
@@ -0,0 +1,28 @@
...
+module DefendantSearchable
+  extend ActiveSupport::Concern
+
+  def defendant
+    @defendant ||= @search.execute.first
+  end
+
+  private
+
+  def defendant_identifier
```
jsugarman, line 12: i thought you renamed this

app/controllers/laa_references_controller.rb
```diff
@@ -1,14 +1,38 @@
...
+  include DefendantSearchable
+
+  before_action :set_defendant_asn_or_nino_if_required,
+                :load_and_authorize_search,
+                :set_link_attempt,
+                :set_defendant_if_required
+
+  add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
+  add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
+
```
jsugarman, line 13: remove line

spec/requests/linking/link_defendant_maat_reference_spec.rb
```diff
@@ -117,7 +118,7 @@
       it 'sends token request' do
         expect(
           a_request(:post, %r{.*/oauth/token})
-        ).to have_been_made.once
+        ).to have_been_made.twice
```
jsugarman, line 121: ?! this implies an issue exists

app/controllers/concerns/defendant_helpers.rb
kmahern: @jsugarman id is asn_or_nino, defendant_id is uuid.

spec/requests/linking/link_defendant_maat_reference_spec.rb
kmahern: @jsugarman is this not because we are now calling load_and_authorize_search to set the defendant in laa_refernces_controller as well as link_laa_reference (so two calls to the adaptor).

jsugarman: as per discussion

PR #217, opened 2020-07-10
Show hearings data currently available via adaptor

jsugarman:
### What
Display available hearing data

#### Ticket
[CBO-1384](https://dsdmoj.atlassian.net/browse/CBO-1384)

### Why
For UAT testing and to initial attempt to meet designs for displaying hearing data


app/controllers/concerns/defendant_searchable.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
kmahern, line 1: @jsugarman, I removed this file in https://github.com/ministryofjustice/laa-court-data-ui/pull/231, we shouldn't need it any more. I don't know whether removing it from your pr would improve code coverage?

app/views/hearings/show.html.haml
```diff
@@ -0,0 +1,75 @@
...
+        = @hearing.court_name || 'not set'
+    %tr.govuk-table__row
+      %td.govuk-table__cell{ class: 'govuk-!-font-weight-bold' }
+        Time listed
+      %td.govuk-table__cell
+        = @hearing_day&.strftime('%H:%M')
+
+%br
+.govuk-heading-m
+  = 'Attendees'
```
naseberry, line 25: - We could remove `%br` as the preceding table element will provide relevant spacing.
- we should add the relevant typography tag to make `.govuk-heading-m` semantically correct
- add `attendees` string to translation

@jsugarman

app/views/hearings/show.html.haml
```diff
@@ -0,0 +1,75 @@
...
+
+%table.govuk-table
+  %caption.govuk-table__caption.govuk-visually-hidden
+    = 'Attendees'
+  %tbody.govuk-table__body
+    %tr.govuk-table__row
+      %td.govuk-table__cell{ class: 'govuk-!-font-weight-bold' }
```
naseberry, line 32: As you don't have a table head, I suggest you change the first `td` of each row to a `th`, like `%th{ class: 'govuk-table__header', scope: 'row' }`.

The `th` font weight is bold, therefore you wont need  the class `govuk-!-font-weight-bold`

@jsugarman

app/views/hearings/show.html.haml
```diff
@@ -0,0 +1,75 @@
...
+        Defendants
```
naseberry, line 33: Inconsistent how you have handled other string using single quotes, plus these should be translated.

@jsugarman

app/views/hearings/show.html.haml
```diff
@@ -0,0 +1,75 @@
...
+    %tr.govuk-table__row
+      %td.govuk-table__cell{ class: 'govuk-!-font-weight-bold' }
+        Hearing type
+      %td.govuk-table__cell
+        = @hearing.hearing_type
+    %tr.govuk-table__row
+      %td.govuk-table__cell{ class: 'govuk-!-font-weight-bold' }
+        Court
+      %td.govuk-table__cell
+        = @hearing.court_name || 'not set'
```
naseberry, line 16: Could we reduce / remove conditional logic from views?

@jsugarman

app/views/hearings/show.html.haml
```diff
@@ -0,0 +1,75 @@
...
+
+%br
+%br
+= govuk_detail('View raw data') do
+  %h3='Hearings'
+  %pre
+    = JSON.pretty_generate(@hearing.attributes)
+  %h3='Hearing events'
+  %pre
+    = JSON.pretty_generate(@hearing.hearing_events.map(&:attributes))
```
naseberry, line 75: Don't forget you have this here, it would be ideal if we could show this based on a user role

@jsugarman

app/controllers/concerns/defendant_searchable.rb
jsugarman: ooh, odd

jsugarman: removed

app/views/hearings/show.html.haml
jsugarman: @naseberry not sure what you mean by point 2, example, but others done

app/views/hearings/show.html.haml
jsugarman: all done now

app/views/hearings/show.html.haml
jsugarman: cool, good idea. I applied this to the defendants table as well

app/views/hearings/show.html.haml
jsugarman: @naseberry this is all still a bit up in the air so would prefer to leave for now. Eventually it would be good to get a decorator pattern in to deal with this logic

app/views/hearings/show.html.haml
jsugarman: have added conditional logic for now based an env var

app/views/hearings/show.html.haml
naseberry: point 2, should it be a heading? Maybe an h2

app/views/hearings/show.html.haml
naseberry: that's ok then :+1:

kmahern:
Not a blocker, especially given time constraints, but would it add any value to add specs for the new hearing and hearing_event court data adaptor resource classes? (I know we have one for the hearing query class)

app/views/hearings/show.html.haml
jsugarman: done now

jsugarman:
> Not a blocker, especially given time constraints, but would it add any value to add specs for the new hearing and hearing_event court data adaptor resource classes? (I know we have one for the hearing query class)

👍  yes, i am adding more specs generally and those were on the list still todo

PR #236, opened 2020-07-30
Multi bug fixes

naseberry:
#### What
Resolve bugs identified during testing

#### Ticket
 - [X] [Unlink reason free text can accept more characters than MLRA can handle](https://dsdmoj.atlassian.net/browse/CBO-1404)
 - [X] [HTML Details does not work on IE11](https://dsdmoj.atlassian.net/browse/CBO-1409)

### DIFF
--- a/app/models/unlink_attempt.rb
+++ b/app/models/unlink_attempt.rb
@@ -8,6 +8,7 @@ class UnlinkAttempt
   validates :username, presence: true
   validates :reason_code, presence: true, inclusion: { in: :valid_reason_codes }
   validates :other_reason_text, presence: { if: :text_required? }
+  validates :other_reason_text, length: { maximum: 500 }
 
   delegate :text_required?, to: :reason, allow_nil: true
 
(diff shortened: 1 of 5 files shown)


app/models/unlink_attempt.rb
```diff
@@ -8,6 +8,10 @@ class UnlinkAttempt
   validates :username, presence: true
   validates :reason_code, presence: true, inclusion: { in: :valid_reason_codes }
   validates :other_reason_text, presence: { if: :text_required? }
+  validates :other_reason_text, length: {
+    maximum: 500,
+    message: I18n.t('activemodel.errors.models.unlink_attempt.attributes.other_reason_text.max_character')
```
jsugarman, line 13: is it not possible to pick this message translation automatically

jsugarman: i think if you just use `.length` and have a message for `length:` you can dispense with the translation here

naseberry: How would I accomplish this?

PR #262, opened 2020-09-02
Change ECR login mechanism

jsugarman:
#### What
Change ECR login mechanism to use get-login-password and --password-stdin

#### Why
More secure and old get-login mechanism is deprecated


naseberry (approved):
👍

PR #277, opened 2020-09-18
Change ECR login mechanism

jsugarman:
Update ECR login mechanism for pushing builds

#### What
Use current preferred login mechanism

#### Why
More secure and old get-login mechanism is deprecated


jsugarman:
@dependabot rebase

PR #289, opened 2020-09-24
Spec devise mailer backgrounding overide

jsugarman:
#### What
Spec devise mailer backgrounding overide

#### Why
should have been spec'd


naseberry (approved):
LGTM

PR #307, opened 2020-10-14
Add prometheus gem and generate metrics endpoint

kmahern:
#### What
Enable prometheus on VCD. This pull request adds and configures the prometheus-client gem. This setups up a new endpoint /metrics (which can be interrogated when running locally with a curl curl localhost:3000/metrics).

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1465

#### Why
The /metrics endpoint can then be scraped by CloudPlatform's Prometheus instance by setting up a separate service-monitor and network policy (see https://github.com/ministryofjustice/cloud-platform-environments/pull/3631). Once these are in place metrics should be available to interrogate in Cloud Platform's prometheus instance (https://prometheus.cloud-platform.service.justice.gov.uk/graph)

#### How
Add and configure prometheus-client gem, as this is the application monitoring tool supported by MOJ cloud platform (following the user guide https://user-guide.cloud-platform.service.justice.gov.uk/#monitoring-applications)


.k8s/dev/service.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   ports:
     - port: 80
-      name: http
+      name: metrics
```
jsugarman, line 11: This seems wrong and likely to break stuff. This is the app/web-app service. Should there not be a separate service for metrics?

.k8s/production/service.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   ports:
     - port: 80
-      name: http
+      name: metrics
```
jsugarman, line 11: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/307/files#r508307575)

.k8s/staging/service.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   ports:
     - port: 80
-      name: http
+      name: metrics
```
jsugarman, line 11: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/307/files#r508307575)

.k8s/uat/service.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   ports:
     - port: 80
-      name: http
+      name: metrics
```
jsugarman, line 11: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/307/files#r508307575)

jsugarman (changes requested):
Not sure this is correct. reusing the web app `service` for metrics seems odd and like it would break stuff. also can you add a fuller description to the PR of what this is doing

kmahern:
I have reverted the name change of port 80 in service.yaml from metrics to http.

jsugarman (approved):
👍  is this still available on `localhost:3000/metrics` endpoint

PR #311, opened 2020-10-19
Fix ruby 2.7 deprecations

jsugarman:
#### What
Fix ruby 2.7 deprecations

#### Why
Ruby 2.7 deprecated use of last argument as keyword parameters.
This fixes all that occur duing test suite run currently.

```
laa-court-data-ui/app/helpers/application_helper.rb:11:
  warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```


kmahern (approved):
lgtm 👍

PR #313, opened 2020-10-19
Extend hearing_events tests to cover properties

jsugarman:
#### What
Add hearing event tests to cover properties
as other PR is for providers.

#### Why
Increase test coverage of this object


kmahern (approved):
👍

PR #340, opened 2020-11-10
Display plea detail and mode of trial reason for defendant

jsugarman:
#### What
Display plea, plea date, and offence mode of trial to defendant view

#### Ticket

[mode of trial reason CBO-1575](https://dsdmoj.atlassian.net/browse/CBO-1575)
[plea details CBO-1588](https://dsdmoj.atlassian.net/browse/CBO-1588)

#### Why
The mode of trial reason and plea details are required information
for caseworkers as it impacts claimable fee values.

#### How

Use amended defedendant endpoint to retrieve these details

see [related PR](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/325)
upon which this functionality depends. However, this could be merged prior
to the adaptor changes without breaking current functionality.

### DIFF
--- a/app/views/defendants/_offences.html.haml
+++ b/app/views/defendants/_offences.html.haml
@@ -15,7 +15,7 @@
         %td.govuk-table__cell
           = offence.title
         %td.govuk-table__cell
-          = offence.plea || t('generic.not_available')
+          = offence.plea_and_date || t('generic.not_available')
         %td.govuk-table__cell
           = "#{offence.mode_of_trial}:"
           %br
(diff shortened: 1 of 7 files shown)


naseberry (approved):
👍  Looks good

naseberry:
Not a blocker: Would it be cleaner to move the conditional logic ` || t('generic.not_available')` from the view to the function in the relevant `lib` folder?
@jsugarman

jsugarman:
> Not a blocker: Would it be cleaner to move the conditional logic ` || t('generic.not_available')` from the view to the function in the relevant `lib` folder?
> @jsugarman

The court data adaptor itself should not store presentation logic but you are correct in that this "not available" logic would be better situation in a "presenter" or "decorator". I did spike something on this related to a previous PR but didi not apply since the decorator gem does not play nicely with non activerecord models (like our JSON response based objects/resources) so we would have to role our own.

it should probably have a separate ticket.

PR #356, opened 2020-11-19
Add metrics container to worker pods

kmahern:
#### What
Fix `Prometheus Exporter, failed to send message Address not available - connect(2) for "localhost" port 9394` error in worker pods

#### Ticket
N/A

#### Why
Currently the worker pods are logging the error above, as they are not able to connect to a prometheus_exporter

#### How
This pr creates a metrics container on worker pods (in the same way as the pods running the application have a metrics container) and runs bundle exec prometheus_exporter in this container. This means that the error above isn't raised and also potentially that we can gather metrics on the worker pods.


jsugarman (approved):
👍  LGTM

PR #357, opened 2020-11-20
Display multiple pleas on an offence

jsugarman:
#### What
Display multiple pleas per offence on defendant details page

#### Ticket

[CBO-1599](https://dsdmoj.atlassian.net/browse/CBO-1599)

depends on:

[CBO-1597](https://dsdmoj.atlassian.net/browse/CBO-1597)

#### Why
Changes of plea are possible and the adaptor is
to be amended to return an array of hashes, where
each hash represents a plea.

#### How
Stub agreed JSON response for offence pleas from the adaptor.

The response is an array of hashes, which is converted
to an array of OpenStruct objects for convenience.
A convenience method plea_list returns an array
of "plea sentences" (e.g. "Guilty on 01/01/2020")
which can then be displayed in the view.

#### Decorator pattern introduced

I have added a decorator/presenter pattern to keep  the presentation logic
out of models/resource-objects while keeping views clean. This pattern is simlar
to CCCDs and can be applied to other classes relatively easily by defining a decorator
(classnameDecorator) and using the new `decorate` or `decorate_all` methods

### DIFF
--- a/app/decorators/offence_decorator.rb
+++ b/app/decorators/offence_decorator.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class OffenceDecorator < BaseDecorator
+  def plea_list
+    return t('generic.not_available') if pleas.blank?
+    return pleas unless pleas.is_a?(Enumerable)
+
+    safe_join(plea_sentences, tag(:br))
+  end
+
+  private
+
+  def plea_sentences
+    sorted_pleas.map { |plea| plea_sentence(plea) }
+  end
+
+  def sorted_pleas
+    pleas.sort_by { |plea| plea&.pleaded_at || Date.new.iso8601 }
+  end
+
+  def plea_sentence(plea)
+    t('offence.plea.sentence',
+      plea: plea.code&.humanize || t('generic.not_available'),
+      pleaded_at: plea.pleaded_at&.to_date&.strftime('%d/%m/%Y') || t('generic.not_available'))
+  end
+end
--- a/app/decorators/offence_decorator.rb
+++ b/app/decorators/offence_decorator.rb
@@ -5,7 +5,7 @@ class OffenceDecorator < BaseDecorator
     return t('generic.not_available') if pleas.blank?
     return pleas unless pleas.is_a?(Enumerable)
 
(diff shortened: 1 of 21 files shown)


app/decorators/offence_decorator.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class OffenceDecorator < BaseDecorator
+  def plea_list
+    return t('generic.not_available') if pleas.blank?
+    return pleas unless pleas.is_a?(Enumerable)
+
+    safe_join(plea_sentences, tag(:br))
```
naseberry, line 8: I thought `tag(:br)` was legacy syntax, should it be `tag.br`

app/helpers/application_helper.rb
```diff
@@ -10,4 +10,19 @@ def service_name
   def l(date, options = {})
     super(date, **options) if date
   end
+
+  # NOTE: implicit decorators assumed to be in app/decorators
+  def decorate(object, decorator_class = nil)
+    decorator_class ||= "#{object.class.to_s.demodulize}Decorator".constantize
+    decorator = decorator_class.new(object, self)
+    yield(decorator) if block_given?
```
naseberry, line 18: Will this work without the ampersand parameter?

app/decorators/offence_decorator.rb
jsugarman: `content_tag(:br, options)` is legacy.  I used `tag(:br)` because it creates one `<br/>` and for some reason `tag.br` creates two `<br/><br/>` - i need to double-check this

app/helpers/application_helper.rb
jsugarman: the `&` is used to pass a reference to a named block to another method usually. `yield(thing)` passes the `thing` to the calling block.

```
decorate(thing) do |decorated_thing_yielded_here|
  decorated_thing_yielded_here.my_decorator_method
end
```

app/decorators/offence_decorator.rb
jsugarman: amended that to `tag.br`, not sure what i did wrong before where it created two

PR #384, opened 2020-12-10
Add new slack alerts webhook linked to the get-paid-alerts slack channel

kmahern:
#### What
Update the SLACK_ALERTS_WEBHOOK secret with a Base64 encoded slack webhook linked to the #get-paid-alerts slack channel.

#### Ticket
N/A

#### Why
It has been requested that VCD alerts are posted in the #get-paid-alerts channel instead of #laa-court-data-ui-dev channel.

#### How
Once the secret is updated we can raise a new issue with Cloud Platforms in order to get the slack alert channel changed.


jsugarman (approved):
👍  should be #**laa**-get-paid-alerts in the PR description I believe but assuming that is what you added

PR #385, opened 2020-12-10
Attempt to quiten nginex-500-error alert

kmahern:
#### What
The nginex-500-error alert is noisy on all environments (especially production). This is an attempt to make it quieter.

#### Ticket
N/A

#### Why
We want the alert to alert us to genuine application issues.

#### How
Change the number of 500 statuses required in a 5 minute period to raise the error from >1 to >5. (I have also  tweaked  * 270 in the expression to * 300 as I am assuming this was a typo for 300 seconds/5 minutes?)


jsugarman (approved):
👍

PR #394, opened 2020-12-18
Apply decorator pattern to hearings and providers

jsugarman:
#### What
Apply decorator pattern to hearings and providers

#### Ticket

N/A

#### Why
Extract presentation layer logic out of hearing 
and provider resource objects and views. Best practice
and consistent with previous work for offences.

#### TODO
- [X] move provider name_and_role to provider decorator
- [X] move hearing.provider_list to hearing decorator
- [x] move hearing.defendant_name presentation to hearing decorator
- [x] move hearing.prosecution_advocate_names presentation to hearing decorator
- [x] move hearing.judge_names presentation to hearing decorator
- [x] test with view specs?!


app/decorators/hearing_decorator.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def defendant_name_list
+    return t('generic.not_available') if defendant_names.blank? || defendant_names.empty?
+
+    safe_join(defendant_names, tag.br)
+  end
+
+  def prosecution_advocate_name_list
+    return t('generic.not_available') \
```
jrmhaig, line 17: I'd prefer this to be on a single line if possible.

app/views/hearings/show.html.haml
```diff
@@ -47,7 +47,7 @@
       %th.govuk-table__header{ scope: 'row' }
         = t('hearings.show.attendees.judges')
       %td.govuk-table__cell
-        = hearing&.judge_names&.join(', ') || t('generic.not_available')
+        = hearing.judge_name_list
```
naseberry, line 50: 🥇

app/decorators/provider_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+    return t('generic.not_available') if name_and_role_blank?
+
+    "#{formatted_name} (#{formatted_role})"
+  end
+
+  private
+
+  def name_and_role_blank?
+    [(respond_to?(:name) && name.blank?),
+     (respond_to?(:role) && role.blank?)].all?
```
jrmhaig, line 14: Is it possible for a provider not to have either of these attributes? From the method you are replacing in lib/court_data_adaptor/resource/provider.rb it looks as though these attributes are presumed to exist. In this case, this check could simply be `name.blank? && role.blank?`.

app/views/hearings/show.html.haml
```diff
@@ -71,7 +70,7 @@
   = govuk_detail('View raw data') do
     %h3='Hearings'
```
naseberry, line 71: All content string be added to locale file.

app/views/prosecution_cases/_raw_response.html.haml
```diff
@@ -0,0 +1,39 @@
+%br
+%br
```
naseberry, line 2: Instead of using line breaks, try using a spacing override classes: https://design-system.service.gov.uk/styles/spacing/

app/decorators/hearing_decorator.rb
jsugarman: @jrmhaig me too, but then it would need to be excluded by rubocop or we allow longer lines in rubocop?!

app/decorators/hearing_decorator.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class HearingDecorator < BaseDecorator
+  def provider_list
+    return t('generic.not_available') if providers.blank? || providers.empty?
```
jrmhaig, line 5: I think you only need `providers.blank?` here as is the same as `.empty?` for arrays and active record relations.

app/decorators/hearing_decorator.rb
jrmhaig: I thought I had written a comment elsewhere about not needing both `.blank?` and `.empty?` but I must not have submitted it. I've added it now (above).

app/decorators/provider_decorator.rb
jsugarman: > Is it possible for a provider not to have either of these attributes? From the method you are replacing in lib/court_data_adaptor/resource/provider.rb it looks as though these attributes are presumed to exist. In this case, this check could simply be `name.blank? && role.blank?`.

I spent a fair bit of time mucking about with this. To get the desired result, which is defensive because of the lack of knowledge of what we can expect from CP, I have used this combination. 

Take a look at the tests for it

app/views/prosecution_cases/_raw_response.html.haml
jsugarman: was just a move of existing content to a partial so a bit of tangent but ok

app/decorators/hearing_decorator.rb
jsugarman: It caused errors without the blank? but that could just be the tests. likewise in the other places. I can revisit to check

app/views/prosecution_cases/_raw_response.html.haml
jsugarman: can you post an example as the `!` causes problems in haml it think

app/decorators/hearing_decorator.rb
jrmhaig: Does it work with `blank?` but not `empty?`?

app/decorators/provider_decorator.rb
jrmhaig: (Please bear with me as I try to understand how it works)

In the `CourtDataAdapter::Resource::Provider` the `name` and `role` are explicitly defined here:

https://github.com/ministryofjustice/laa-court-data-ui/blob/d61abdc6f88778f2c87e4fdc25970f7fcbaa665d/lib/court_data_adaptor/resource/provider.rb#L10-L11

So those attributes will always exist, even if they are not returned from the CP, but they may be `nil`.

As it says in the [Json Api Client documentation](https://github.com/JsonApiClient/json_api_client#schema) these definitions are optional but when they are present it handles default values and type casting.

app/decorators/hearing_decorator.rb
jsugarman: I see yes 👍

app/decorators/provider_decorator.rb
jsugarman: Yes,  because they are explicitly defined in the resource object these attributes/methods will always be available and be at least nil and therefore the `respond_to` is not needed.

However, these [tests were done in isolation](https://github.com/ministryofjustice/laa-court-data-ui/pull/394/files#diff-16f7eab620be5bf8db87c7798b3b0fab45ab6690e54ae8bb1c3ae6f26c45b222R52-R62) as part of presenter logic alone and so were made to handle the possibility that they may raise an error. 

The tests and therefore the checks may be excessive but it is intended to eventually move the court data adaptor lib to a gem, in which case the presenter logic is all we have. I wonder if we may need to be more careful with other attributes too.

For simplicity I could remove these tests and checks for now?!

app/views/prosecution_cases/_raw_response.html.haml
naseberry: Because `!` causes issues you have to wrap it within braces `{class: 'govuk-!-margin-top-2'}`

app/views/hearings/show.html.haml
jsugarman: done

app/views/prosecution_cases/_raw_response.html.haml
jsugarman: done now

app/decorators/hearing_decorator.rb
jsugarman: done

app/decorators/provider_decorator.rb
jrmhaig: I can see that, when moving the adaptor to a gem, the checks that `name` and `role` exist will be separated from the presenter. However, isn't it the purpose of the adaptor to ensure that a consistent public interface is presented regardless of what appears upstream? I suppose the question is whether it is likely that these attributes will be removed from the gems interface in the future. If this happens I would expect them to report a deprecation warning in (at least) one version, rather than just being removed.

app/decorators/hearing_decorator.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class HearingDecorator < BaseDecorator
+  def provider_list
+    return t('generic.not_available') if providers.blank?
+
+    safe_join(provider_sentences, tag.br)
```
kmahern, line 7: We should agree a consistent approach for separators, have asked Dan Frampton if he has an opinion.

jsugarman: tangent at this point. separate PR

app/decorators/provider_decorator.rb
jsugarman: I will remove for now as is obviously controversial and excessive in any event

jsugarman: done now

app/decorators/hearing_decorator.rb
kmahern: 👍

kmahern (approved):
lgtm 👍

jrmhaig:
👍

jrmhaig (approved):

PR #409, opened 2021-01-04
Update JS & SCSS linting

naseberry:
#### What
Update JS & SCSS linting

#### Ticket
[Update JS & SCSS linting](https://dsdmoj.atlassian.net/browse/CBO-1636)

#### Why
To enforce consistent linting tooling & strategy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,7 +126,7 @@ commands:
     steps:
       - run:
           name: Run JS-lint
-          command: yarn run lint-js
+          command: yarn run validate:js
       - run:
           name: Run Stylelint
           command: yarn run validate:scss
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,8 +128,8 @@ commands:
           name: Run JS-lint
           command: yarn run lint-js
       - run:
-          name: Run Sass-lint
-          command: yarn run lint-sass
+          name: Run Stylelint
+          command: yarn run validate:scss
 
   brakeman:
     steps:
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -22,3 +22,21 @@ if [[ $? -ne 0 ]]; then
 else
   printf "none found!"
 fi
+
+
+################################################################################
+# JavaScript linting using standardJS
+#
+
+printf '\nLinting JavaScript files...\n'
+
+yarn run validate:js
+
+
+################################################################################
+# SCSS linting using styleLint
+#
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -122,15 +122,6 @@ commands:
...
 
-  linter:
-    steps:
-      - run:
-          name: Run JS-lint
-          command: yarn run lint-js
-      - run:
-          name: Run Sass-lint
-          command: yarn run lint-sass
-
```
jsugarman, line 133: I guess we removed JS/CSS lint step in CCCD too?! Personally I quite like doing it in the CI too because:

1. its quick anyway
2. devs may not have the the git-hook symlinked, so provides a double-check
3. If we change CI tool (github actions in the future) this provides a record of what is being checked.

jsugarman (commented):
please see my comment

.circleci/config.yml
naseberry: > I guess we removed JS/CSS lint step in CCCD too?! Personally I quite like doing it in the CI too because:
> 
> 1. its quick anyway
> 2. devs may not have the the git-hook symlinked, so provides a double-check
> 3. If we change CI tool (github actions in the future) this provides a record of what is being checked.

This was never implemented in the CI for CCCD.

1. it is faster locally as it doesnt have to run all the steps that CI does.
2. If repo cloned, it should all 'just work'
3. This works regardless of the CI tool used as it will always run locally when a dev runs `$ git commit`

@jsugarman

jsugarman: > 2. If repo cloned, it should all 'just work'

I don't think it will, as [per instructions here](https://github.com/ministryofjustice/laa-court-data-ui/blob/master/.git-hooks/pre-commit#L3-L6) the dev needs to symlink this hook to their local .git/hooks/pre-commit hook. That is certainly how i set it up. In addition a dev may not want or be able to run the git-crypt aspect of this unless they have access to the secrets. I am not sure what happens if you do not have git-crypt access TBH

> 1. it is faster locally as it doesnt have to run all the steps that CI does.

no doubt, but I feel CI is the central place for linting and testing (rubocop is also a linter)

> 3. This works regardless of the CI tool used as it will always run locally when a dev runs $ git commit

true but because of 2. above it could get "lost" in any transition.

Fundamentally i see the git hook as a useful option to use, but CI should really be the central place for all linting and testing.

jsugarman (approved):
👍 LGTM

PR #416, opened 2021-01-08
Specify version of docker to use

jsugarman:
#### What
Specify version of docker to use

#### Why
Started receiving this during app container build:
```
[3/4] Linking dependencies...
error An unexpected error occurred: "EPERM: operation not permitted, copyfile '/usr/local/share/.cache/yarn/v6/...blahblahblah".
info If you think this is a bug, please open a bug report with the information provided in "/usr/src/app/yarn-error.log".
```

See this [related discussion](https://discuss.circleci.com/t/docker-build-fails-with-nonsensical-eperm-operation-not-permitted-copyfile/37364)

#### How
specify version of remote docker as described here

https://circleci.com/docs/2.0/building-docker-images/#docker-version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,7 +189,8 @@ commands:
         type: string
     steps:
       - checkout
-      - setup_remote_docker
+      - setup_remote_docker:
+          version: 19.03.13
       - run:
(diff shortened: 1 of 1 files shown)


naseberry (approved):
💯

PR #417, opened 2021-01-11
Hide mode_of_trial_reasons with certain codes

kmahern:
#### What
Apply mappings for common platform mode of trial reasons

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1633

#### Why
So that mode_of_trial_reason is hidden for certain associated codes (see https://docs.google.com/spreadsheets/d/1-gkLXxS7DqEPDf5tdMkc3c6_Rtrq2G5xIH6OMSvNyYY/edit#gid=1021818802) and to allow for possible future mode_of_trial_reason mappings to be defined.

#### How
Added MODE_OF_TRIAL_REASON_MAPPINGS, currently this allows for mode_of_trial_reason to be hidden for certain associated codes, however, it can be extended in future to allow mode_of_trial_reasons.description to be mapped to different values as required (e.g. 'Directed' and 'Elected' etc.)

### DIFF
--- a/app/decorators/offence_decorator.rb
+++ b/app/decorators/offence_decorator.rb
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 
 class OffenceDecorator < BaseDecorator
+  # Allow for future mot reason mappings; e.g. to 'Elected' and 'Directed'
+  MODE_OF_TRIAL_REASON_MAPPINGS = {
+    1 => '',
+    2 => '',
+    6 => ''
+  }.freeze
+
   def plea_list
     return t('generic.not_available') if pleas.blank?
     return pleas unless pleas.is_a?(Enumerable)
@@ -12,7 +19,7 @@ class OffenceDecorator < BaseDecorator
     return t('generic.not_available') if mode_of_trial_reasons.blank?
     return mode_of_trial_reasons unless mode_of_trial_reasons.is_a?(Enumerable)
 
-    safe_join(mode_of_trial_reason_descriptions, tag.br)
+    safe_join(mode_of_trial_reason_descriptions.reject(&:empty?), tag.br)
   end
 
   private
@@ -32,6 +39,14 @@ class OffenceDecorator < BaseDecorator
   end
 
   def mode_of_trial_reason_descriptions
-    mode_of_trial_reasons.map { |reason| reason.description || t('generic.not_available') }
+    mode_of_trial_reasons.map { |reason| mode_of_trial_reason_description(reason) }
+  end
+
+  def mode_of_trial_reason_description(reason)
+    if MODE_OF_TRIAL_REASON_MAPPINGS.key?(reason.code.to_i)
+      MODE_OF_TRIAL_REASON_MAPPINGS[reason.code.to_i]
+    else
+      reason.description || t('generic.not_available')
+    end
   end
 end
--- a/app/decorators/offence_decorator.rb
+++ b/app/decorators/offence_decorator.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class OffenceDecorator < BaseDecorator
-  # Allow for future mot reason mappings; e.g. to 'Elected' and 'Directed'
+  # TODO: mot reason text mappings for 'Elected' and 'Directed'
   MODE_OF_TRIAL_REASON_MAPPINGS = {
     1 => nil,
     2 => nil,
--- a/app/decorators/offence_decorator.rb
+++ b/app/decorators/offence_decorator.rb
@@ -3,9 +3,9 @@
 class OffenceDecorator < BaseDecorator
   # Allow for future mot reason mappings; e.g. to 'Elected' and 'Directed'
   MODE_OF_TRIAL_REASON_MAPPINGS = {
-    1 => '',
-    2 => '',
-    6 => ''
+    1 => nil,
+    2 => nil,
+    6 => nil
   }.freeze
 
   def plea_list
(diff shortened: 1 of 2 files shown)


spec/decorators/offence_decorator_spec.rb
```diff
@@ -185,5 +187,44 @@
...
+          let(:mode_of_trial_reason_array) do
+            [{ code: code,
+               description: 'Defendant elects trial by jury' }]
+          end
+
+          it { is_expected.to eql('') }
+        end
+      end
+
+      context 'when more than one reason exists, one of which should be hidden' do
```
jrmhaig, line 207: If I understand the codes correctly then this description and the next, on line 218, are the wrong way round.

app/decorators/offence_decorator.rb
```diff
@@ -32,6 +39,14 @@ def plea_sentence(plea)
...
-    mode_of_trial_reasons.map { |reason| reason.description || t('generic.not_available') }
+    mode_of_trial_reasons.map { |reason| mode_of_trial_reason_description(reason) }
+  end
+
+  def mode_of_trial_reason_description(reason)
+    if MODE_OF_TRIAL_REASON_MAPPINGS.key?(reason.code.to_i)
+      MODE_OF_TRIAL_REASON_MAPPINGS[reason.code.to_i]
+    else
+      reason.description || t('generic.not_available')
+    end
```
jrmhaig, line 50: You may be able to replace this with:

```ruby
MODE_OF_TRIAL_REASON_MAPPINGS[reason.code.to_i] || reason.description || t('generic.not_available')
```

(but I haven't checked).

jrmhaig (approved):
:+1: A couple of minor comments but looks good.

app/decorators/offence_decorator.rb
```diff
@@ -12,7 +19,7 @@ def mode_of_trial_reason_list
     return t('generic.not_available') if mode_of_trial_reasons.blank?
     return mode_of_trial_reasons unless mode_of_trial_reasons.is_a?(Enumerable)
 
-    safe_join(mode_of_trial_reason_descriptions, tag.br)
+    safe_join(mode_of_trial_reason_descriptions.reject(&:empty?), tag.br)
```
jsugarman, line 22: will [compact](https://ruby-doc.org/core-2.4.1/Array.html#method-i-compact) work here instead of `empty?`

```suggestion
    safe_join(mode_of_trial_reason_descriptions.compact, tag.br)
```

spec/decorators/offence_decorator_spec.rb
```diff
@@ -185,5 +187,44 @@
...
+    context 'when the mode of trial reason code means the description should be hidden' do
+      before do
+        allow(offence).to receive(:mode_of_trial_reasons).and_return(mot_reason_ostruct_collection)
+      end
+
+      [1, 2, 6].each do |code|
+        context "when exactly one reason exists, with code #{code}" do
+          let(:mode_of_trial_reason_array) do
+            [{ code: code,
+               description: 'Defendant elects trial by jury' }]
```
jsugarman, line 200: for the tests i'd use either a real code/description pair that are to be hidden, or a description something like 'to be hidden'. I think "Defendant elects trial by jury" is one that is to be shown

```suggestion
               description: 'Indictable-only offence' }]
```

spec/decorators/offence_decorator_spec.rb
```diff
@@ -185,5 +187,44 @@
...
+          end
+
+          it { is_expected.to eql('') }
+        end
+      end
+
+      context 'when more than one reason exists, one of which should be hidden' do
+        let(:mode_of_trial_reason_array) do
+          [{ code: '2',
```
jsugarman, line 209: nit, but similar to above but using real data 
```suggestion
          [{ code: '4',
```

spec/decorators/offence_decorator_spec.rb
jsugarman: yes, the context descriptions look like they need swapping around with line `218`

app/decorators/offence_decorator.rb
jsugarman: no it won't actually. you could use `nil` instead of empty string and then use compact

app/decorators/offence_decorator.rb
jsugarman: or possibly 
```
MODE_OF_TRIAL_REASON_MAPPINGS.fetch(reason.code.to_i, reason.description) || t('generic.not_available')
```

spec/decorators/offence_decorator_spec.rb
kmahern: 👍 have amended/swapped the context descriptions

app/decorators/offence_decorator.rb
kmahern: Neither of these suggestions quite worked but have amended to `MODE_OF_TRIAL_REASON_MAPPINGS.fetch(reason.code.to_i, reason.description || t('generic.not_available'))` which is similar.

app/decorators/offence_decorator.rb
kmahern: 👍 have amended to use nil/compact

spec/decorators/offence_decorator_spec.rb
kmahern: 👍 have amended spec to use 'description to be hidden'

spec/decorators/offence_decorator_spec.rb
kmahern: 👍 have amended spec to use realistic data

app/decorators/offence_decorator.rb
jsugarman: cool, i think nil is better because it can also be used as a `falsey`

app/decorators/offence_decorator.rb
```diff
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 
 class OffenceDecorator < BaseDecorator
+  # Allow for future mot reason mappings; e.g. to 'Elected' and 'Directed'
```
jsugarman, line 4: minor nit, 
```suggestion
  # TODO: mot reason text mappings for 'Elected' and 'Directed'
```

jsugarman (approved):
👍 Looks good

PR #418, opened 2021-01-11
Display cracked ineffective trial details

jsugarman:
#### What
Display cracked ineffective trial details

#### Ticket

[CBO-1606](https://dsdmoj.atlassian.net/browse/CBO-1606)

#### Why
Without sufficient information in VCD on what events lead to a trial cracking, caseworkers may not be able to authorise cracked trial bills on CP cases. Cracked Trials make up a sizeable % of the bills so its vital we ensure the necessary info is available for assessment.

There are 59 cracked ineffective reason codes. After doing analysis on this data we now know that a proportion of these can relate to various case types and the majority of them are actually reasons why a specific individual hearing does not take place (ineffective and vacated).

#### Questions/issues

- crackedIneffectiveTrial details are on the hearing level and have no defendant ID - but cracked results can be (are always/mostly?) defendant specific. This would require a change to Common platform. Since only 1% of cases have more than one defendant it is an edge case.

--------

#### TODO (wip)

 - [x] prosecution case page - cracked ineffective trial details
       for "cracked" types and the 6 "vacated" types
 - [x] hearing page - any cracked ineffective trial detail
 - [x] prosecution case page - handle multiple case "cracked" results


app/helpers/application_helper.rb
```diff
@@ -15,7 +15,10 @@ def l(date, options = {})
   def decorate(object, decorator_class = nil)
     decorator_class ||= "#{object.class.to_s.demodulize}Decorator".constantize
     decorator = decorator_class.new(object, self)
-    yield(decorator) if block_given?
+
+    yield(decorator) if block_given? && decorator.present?
+    return nil if decorator.blank?
```
jrmhaig, line 20: Possibly my misunderstanding.

Isn't the purpose of the Null Object Pattern so that you do not need to do these checks (`decorator.present?` and `decorator.blank?`), and the null instance can be used as normal?

jsugarman: Only the helper relies on the NilClassDecorator to explicitly return a `nil`. In the views a decorated object which was `nil` will therefore remain `nil`

Sadly a nil decorator object is not `nil`   - it is an instance of `NilClassDecorator(nil)` which is always truthy - only the `nil` object itself is "falsey"

```
NilClassDecorator(nil) => truthy
BUT
NilClassDecorator(nil).nil? => true
NilClassDecorator(nil).blank? => true
NilClassDecorator(nil).present? => false
```

jsugarman: quite a useful [summary of ruby nil here](https://www.rubyguides.com/2018/01/ruby-nil/)

.rubocop.yml
```diff
@@ -103,6 +103,12 @@ RSpec/AnyInstance:
 RSpec/Capybara/FeatureMethods:
   EnabledMethods: ['feature', 'scenario']
 
+RSpec/ContextWording:
+  Prefixes:
+    - when
+    - with
+    - without
```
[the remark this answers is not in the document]
jsugarman: They are the defaults  yes. i was going to add to them (with `for`) but did not in the end. I quite like having them explicitly defined, but can remove.

kmahern (approved):
Lgtm other than code climate issue 👍 .

jrmhaig (approved):
👍 

Looks like the Code Climate problem is fixed now.

naseberry:
Nothing to comment on from the FE perspective.
Note: In general, we need to reduce the logic within views and consider discussing/implementing a table helper

PR #429, opened 2021-01-21
Prevent nilclass errors

jsugarman:
#### What
Fix nil class error. will spec after the fact
but this is a prod issue

#### Why
sentry errors show this against UAT
```
ActionView::Template::Error ProsecutionCasesController#show
error
undefined method `cracked?' for nil:NilClass
```


kmahern (approved):
👍

jrmhaig (approved):

PR #434, opened 2021-01-22
Spec successful rendering of cracked ineffective trial partial

jsugarman:
#### What
Spec the common situation where a mixture of hearings
with and without a cracked_ineffective_trial relation
exists on a prosecution_case

#### Why
Common situation which raised errors on UAT. Have not
seen such errors on production as yet, but would expect
them.

The safe navigation operator was previously used to
hotfix this scenario.


jrmhaig (approved):
:+1: Looks OK

PR #437, opened 2021-01-26
Migrate to sentry ruby

kmahern:
#### What
Migrate from sentry-raven gem to sentry-ruby gem.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1674

#### Why
sentry-raven is no longer supported

#### How
Added sentry-ruby gem. Also included config for breadcrumb tracking. I have tested that exceptions are still captured on Sentry on laa-court-data-ui-uat.


config/initializers/raven.rb
```diff
@@ -1,11 +0,0 @@
-# frozen_string_literal: true
-
-require 'raven'
-
-if Rails.env.eql?('production') && ENV['SENTRY_DSN'].present?
-  Raven.configure do |config|
-    config.dsn = ENV['SENTRY_DSN']
-    config.environments = Rails.env
-    config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
```
jrmhaig, line 9: This ensures that certain fields (eg password) do not appear in the fields. Is it part of the default configuration for `sentry-rails`?

kmahern: According to this https://github.com/getsentry/sentry-ruby sentry-ruby does not report personably identifiable information (including the request body) by default. I believe this would stop passwords from appearing? config.sanitize_fields was removed (https://docs.sentry.io/platforms/ruby/migration/)

Gemfile
```diff
@@ -25,7 +25,9 @@ gem 'prometheus_exporter'
 gem 'puma', '~> 5.1'
 gem 'rails', '~> 6.1.0'
 gem 'redis', '~> 4.2.5'
-gem 'sentry-raven'
+gem 'sentry-rails'
+gem 'sentry-ruby'
+gem 'sentry-sidekiq'
```
jrmhaig, line 30: I think these need to have their versions pinned, to be consistent with the rest of the `Gemfile`. Possibly:

```ruby
gem 'sentry-rails', '~> 4.1.5'
gem 'sentry-sidekiq', '~> 4.1.2'
```

Also (not a major issue) `sentry-ruby` is a dependency of `sentry-rails` and `sentry-sidekiq` so it doesn't need to be explicitly listed.

kmahern: 👍 I have pinned the gemfile versions and removed sentry-ruby from the Gemfile.

jrmhaig (approved):
👍 Looks good

PR #443, opened 2021-01-28
Add team devs as default dependabot reviewers

jsugarman:
#### What
Add team devs as default dependabot reviewers

#### Ticket

n/a

#### Why
Any team dev can review and approve dependabot
PR.s this automates the assignment of devs.

Note: we should move to version 2 of the
dependabot githubnative config. see
cccd for example.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -10,6 +10,8 @@ update_configs:
     # pull requests
     default_reviewers:
       - "jsugarman"
+      - "kmahern"
+      - "jrmhaig"
 
   # Keep package.json (& lockfiles) up to date as soon as
   # new versions are published to the npm registry


naseberry:
I was thinking, instead of the individuals, whom we would have to remove from code when there is a change to the team structure, why not add the github team 'laa-get-paid'?

jsugarman:
> I was thinking, instead of the individuals, whom we would have to remove from code when there is a change to the team structure, why not add the github team 'laa-get-paid'?

I don't think that allows you to easily just add your individual review - but have not tried it. The idea is to a) ping the individuals to notify them they can review (if they have notifications enabled) and b) remove the need to assign yourself to the PR and then approve. We already have this on CCCD incidentally

I its pretty easy to amend to remove a dev if they leave so i don't see that as a big issue

naseberry (approved):

PR #447, opened 2021-01-30
Handle unexpected formats for not_found responses

jsugarman:
#### What
Handle unexpected formats for not_found responses

#### Why
Prevent sentry error clutter from requests to non-existent paths/pages
with unhandled formats as well.

Example error:

```
ActionView::MissingTemplate
  Missing template errors/not_found, application/not_found with {:locale=>[:en], :formats=>[:text], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :haml]}
```

The missing template error results from the format (e.g. `:formats=>[:text]`)
itself. Such requests appear to be be vulnerability scans. As such they are not
something we need cluttering up sentry - they are recorded
by our prometheus monitoring and alerting setup in any event.

#### How
Keep responding to `html` requests with the errors#not_found template
as usual, but all other formats can respond with plain text.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -11,7 +11,10 @@ class ErrorsController < ApplicationController
   end
 
   def not_found
-    render status: :not_found
+    respond_to do |format|
+      format.html { render status: :not_found }
+      format.all { render plain: 'Not found', status: :not_found }
+    end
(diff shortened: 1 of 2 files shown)


jrmhaig (approved):
:+1: Looks OK

PR #450, opened 2021-02-01
Bump bootsnap 1.6.0 --> 1.7.0

jsugarman:
Getting build errors as below. bootsnap is
the dependency that use msgpack and it has released
a fix today thankfully.

```
Your bundle is locked to msgpack (1.4.1-java), but that version could not be
found in any of the sources listed in your Gemfile. If you haven't changed
sources, that means the author of msgpack (1.4.1-java) has removed it. You'll
need to update your bundle to a version other than msgpack (1.4.1-java) that
hasn't been removed in order to install.
```

Bootsnap [changelog](https://github.com/Shopify/bootsnap/blob/master/CHANGELOG.md)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby '2.7.0'
 
 # Reduces boot times through caching; required in config/boot.rb
-gem 'bootsnap', '>= 1.4.2', require: false
+gem 'bootsnap', '>= 1.7.0', require: false
 
 gem 'breadcrumbs_on_rails'
 gem 'cancancan'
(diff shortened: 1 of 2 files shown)


kmahern (approved):
👍

jrmhaig (approved):
:+1: Looks OK.

I'm not sure why the bootsnap version was preventing msgpack from upgrading. The requirement was '~> 1.0' so version 1.4.2 should have been fine.

jsugarman:
> 👍 Looks OK.
> 
> I'm not sure why the bootsnap version was preventing msgpack from upgrading. The requirement was '~> 1.0' so version 1.4.2 should have been fine.

🤷  but the gemfile lock was using 1.4.1 - perhaps a bundle update msgpack would have done - but i assume bootsnap 
changed their dependencies as part of 1.7.0

jsugarman:
it does not look like bootsnap bumped their dependency so i guess a bundle update bootsnap 1.6.0 might have done the trick

PR #453, opened 2021-02-02
[CBO-1428] Display single hearing day logs

jsugarman:
#### What
Display a single hearing days worth of hearing events

#### Ticket

[CBO-1428](https://dsdmoj.atlassian.net/browse/CBO-1428)

#### Why

To meet current designs:

##### User story
When a user navigates to the hearing summaries page
And the user sees multiple dates for the same hearing
And the user selects a specific day of the hearing
Then they are taken to the hearing page
And they can see the hearing day at the top of the page (currently this date cannot be relied on to be the one they selected)
And they can see only events for that specific hearing day


naseberry:
Views are ok, however it would be nice to reduce the logic in them.
I'll leave approval to the others.

jsugarman:
> Views are ok, however it would be nice to reduce the logic in them.
> I'll leave approval to the others.

thanks, what view logic are you referring to? you can comment on specific bits if you prefer. Much of this work has been removing logic from the views

naseberry:
> > Views are ok, however it would be nice to reduce the logic in them.
> > I'll leave approval to the others.
> 
> thanks, what view logic are you referring to? you can comment on specific bits if you prefer. Much of this work has been removing logic from the views

For example where we are using the OR operators `||`
The method for `court_name` should have the logic and return a value of the actual court name or 'not available', thereby removing the OR operator from the view.

spec/lib/court_data_adaptor/resource/defendant_spec.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   it { is_expected.to respond_to(:prosecution_case_reference, :prosecution_case_reference=) }
-  it { is_expected.to respond_to(:name) }
+  it { is_expected.to respond_to(:id, :name, :date_of_birth, :maat_reference) }
```
kmahern, line 18: Not a blocker: if we are adding in these respond_to matchers for lib/court_data_adaptor/resource/defendant.rb should we also add those associated with linking/unlinking (i.e. user_name,  unlink_reason_code, unlink_reason_text)

kmahern (approved):
One small non-blocking comment otherwise lgtm 👍 (I have had a look to see if I could work out why it was complaining about code coverage by running the specs and looking at  /coverage but couldn't see why).

jrmhaig (approved):
:+1: Looks good

I had a couple of questions but managed to work them out as I was writing them up. :-)

spec/lib/court_data_adaptor/resource/defendant_spec.rb
jsugarman: makes sense, especially since those properties are needed for the update/linking of a defendant

jsugarman:
> > > Views are ok, however it would be nice to reduce the logic in them.
> > > I'll leave approval to the others.
> > 
> > 
> > thanks, what view logic are you referring to? you can comment on specific bits if you prefer. Much of this work has been removing logic from the views
> 
> For example where we are using the OR operators `||`
> The method for `court_name` should have the logic and return a value of the actual court name or 'not available', thereby removing the OR operator from the view.

True, won't do that in this PR as the refactoring was getting to be the majority of the PR anyway. Feel free to PR such changes yourself or I will try and put one together later today

spec/lib/court_data_adaptor/resource/defendant_spec.rb
jsugarman: done

PR #456, opened 2021-02-04
Change names search to handle extra whitespace

jsugarman:
#### What
squish names for search to ensure extra whitespace
has no impact

#### Why
Additional whitespace is being added to the
the name coming from the adaptor, making
comparison fail.

I have [raised an issue with the adaptor](https://github.com/ministryofjustice/laa-court-data-adaptor/issues/395)
and a [fix PR for it](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/397) but we can handle this our side to for assurance.

#### How
By removing whitespace from ends and between multiple names
and doing a case insensitive search against the adaptor
results we can remove search failures resulting from whitespace

### DIFF
--- a/lib/court_data_adaptor/query/defendant/by_name.rb
+++ b/lib/court_data_adaptor/query/defendant/by_name.rb
@@ -23,7 +23,7 @@ module CourtDataAdaptor
         private
 
         def name
-          term.strip
+          term.squish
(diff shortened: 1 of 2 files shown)


naseberry (approved):
lgtm

PR #458, opened 2021-02-05
Change header for hearings page

jsugarman:
#### What
Change header and breadcrumb for hearings page to "Hearing day {date}"

#### Ticket

[CBO-1417](https://dsdmoj.atlassian.net/browse/CBO-1417)

#### Why
Designer wants to make it clear to users that they are
only viewing a single hearing days event logs
for the hearing.


naseberry (approved):
lgtm

jsugarman:
I guess i should change the caption of the table to match the header too - i.e. "hearing day" instead of "hearing" - from an accesibility point of view @naseberry

PR #463, opened 2021-02-09
Do not display a link to the hearing_day where the hearing_day is in …

kmahern:
#### What
Do not display a link to the hearing_day where the hearing_day is in the future.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1694

#### Why
Currently all hearings are being handled the same – the date is a link that navigates to the specific hearing day. This means future hearings are not visually distinguished from those that have happened and contain events etc, and also that clicking a link causes an error, which is unexpected (no warning) and frustrating.

#### How
Do not display link if hearing_day > Date.today

### DIFF
--- a/app/views/prosecution_cases/_hearing_summaries.html.haml
+++ b/app/views/prosecution_cases/_hearing_summaries.html.haml
@@ -13,7 +13,12 @@
       - hearing.hearing_days.map(&:to_date).sort.each do |hearing_day|
         %tr.govuk-table__row
           %td.govuk-table__cell
-            = link_to hearing_day.to_date.strftime('%d/%m/%Y'), hearing_path(id: hearing.id, urn: prosecution_case.prosecution_case_reference, hearing_day: hearing_day.to_date.strftime('%d/%m/%Y')), class: 'govuk-link'
+            - if hearing_day > Date.today
+              = hearing_day.to_date.strftime('%d/%m/%Y')
+              %br
+              = 'Scheduled'
+            - else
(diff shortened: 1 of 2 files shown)


app/views/prosecution_cases/_hearing_summaries.html.haml
```diff
@@ -13,7 +13,12 @@
       - hearing.hearing_days.map(&:to_date).sort.each do |hearing_day|
         %tr.govuk-table__row
           %td.govuk-table__cell
-            = link_to hearing_day.to_date.strftime('%d/%m/%Y'), hearing_path(id: hearing.id, urn: prosecution_case.prosecution_case_reference, hearing_day: hearing_day.to_date.strftime('%d/%m/%Y')), class: 'govuk-link'
+            - if hearing_day > Date.today
```
jsugarman, line 16: better to use rails `Date.current` - see [docs for difference](https://api.rubyonrails.org/v5.1.7/classes/Date.html#method-c-current)

Also, not ideal to put conditional logic in the view. To handle it could be moved to a decorator. However, as part of pagination I am looking at amending this so probably
not worth it at the moment. 

on the issue of pagination I can see this is going to cause issues as well as it effectively means we have to skip over some days. For that reason I am not sure we should be doing this now at all unless this is an established feature of Common platform - what is the actual harm in linking to a future hearing anyway? Won't it just will give any available details for the hearing/day?!

jsugarman (commented):
Please see my [comment](https://github.com/ministryofjustice/laa-court-data-ui/pull/463/files#r572810495)

app/views/prosecution_cases/_hearing_summaries.html.haml
kmahern: 👍

kmahern:
Closing this pr as this issue will be looked at as part of pagination (see comment above).

PR #464, opened 2021-02-09
Add missing govuk-link class to links without it

jsugarman:
#### What
Add missing govuk-link class to links without it

#### Ticket

n/a

#### Why
Noticed there were some missing links. I believe the
the devise mailers are not used to it should not matter
but have amended for completeness.


naseberry (approved):
lgtm

PR #471, opened 2021-02-12
Add hearing day pagination

jsugarman:
#### What
Add next and previous hearing day and hearing pagination

#### Ticket

[CBO-1417 hearing day pagination](https://dsdmoj.atlassian.net/browse/CBO-1417)

#### Why
Usability testing showed that caseworkers (especially AGFS)
need to view the detail of all hearing days, especially for
trials.

Adding next/previous links made this quicker and easier,
where as previously they could only navigate via the list
of hearings, which was inefficient and frustrating for
longer cases.

#### How
Unfortunately the usual pagination gems (will_paginate
and kaminari) are not going to work because:

a) we are consuming an API, not Active record models

b) although we are using json_api_client gem, that
   follows the jsonapi specification and is [supposed
   to be compatible with kaminari and will_paginate](https://jsonapi.org/format/#fetching-pagination)
   the Court data adaptor api itself does not implement the [jsonapi pagination specification](https://jsonapi.org/format/#fetching-pagination).

c) The nature of data coming from common platform and then from the adaptor
   does not lend itself to "out of the box" pagination solutions and
   a custom solution is likely to be more flexible while we progress our
   understanding and requirements.

--------

#### TODO (wip)

 - [ ] speed improvement options: attempt to pass a cached prosecution case;
       pass a PageItem set; use caching of api responses
 - [ ] write issue on CDA for hearing_day filter on hearing endpoint
 - [ ] write issue on CDA for jsnoapi pagination
 - [ ] write issue on CDA for jsonapi sorting

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -13,9 +13,9 @@ class HearingsController < ApplicationController
                  (proc { |v| v.prosecution_case_path(v.controller.prosecution_case_reference) })
 
   def show
-    add_breadcrumb "#{t('generic.hearing_day')} #{@hearing_day&.strftime('%d/%m/%Y')}", ''
+    add_breadcrumb "#{t('generic.hearing_day')} #{hearing_day&.strftime('%d/%m/%Y')}", ''
 
-    return if @hearing
+    return if hearing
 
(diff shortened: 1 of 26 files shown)


jsugarman:
closed in favour of #475

PR #472, opened 2021-02-12
Configure vcr hook to match on path and query

jsugarman:
#### What
specs tagged with :vcr will only
match requests to recordings based on
the path and query of the request, not
the domain.

#### Why
Allows cassettes recorded against one
adaptor domain (or local) to function
as expected for feature tests, irrespective
of the app configuration which defines which
adaptor domain to make requests to.

### DIFF
--- a/spec/support/vcr.rb
+++ b/spec/support/vcr.rb
@@ -63,7 +63,7 @@ RSpec.configure do |config|
   config.around(:each, :vcr) do |example|
     if VCR.turned_on?
       cassette = cassette_name(example)
-      VCR.use_cassette(cassette, record: :new_episodes) do
+      VCR.use_cassette(cassette, record: :new_episodes, match_requests_on: %i[method path query]) do
         example.run
       end
     else


spec/support/vcr.rb
```diff
@@ -63,7 +63,7 @@
   config.around(:each, :vcr) do |example|
     if VCR.turned_on?
       cassette = cassette_name(example)
-      VCR.use_cassette(cassette, record: :new_episodes) do
+      VCR.use_cassette(cassette, record: :new_episodes, match_requests_on: %i[path query]) do
```
jrmhaig, line 66: My only question is whether you want to include `:method` in this too. As it stands it will use the same cassette for a GET or a POST to the same path and query. (Of course, a POST would not have a query but might have a body)

jsugarman: i will give it a go and if it works fine with existing features then i will add it

jsugarman: @jrmhaig that worked and done now

jrmhaig (approved):
:+1: Looks good

PR #473, opened 2021-02-12
Extract common rspec `let` api_url to a helper

jsugarman:
#### What
move the api_url to a spec support helper

#### Why
Its use in quite a few places


jrmhaig (approved):
:+1:

PR #475, opened 2021-02-12
[CBO-1417] hearing day pagination

jsugarman:
#### What
Add next and previous hearing day and hearing pagination

#### Ticket

[CBO-1417 hearing day pagination](https://dsdmoj.atlassian.net/browse/CBO-1417)

#### Why
Usability testing showed that caseworkers (especially AGFS)
need to view the detail of all hearing days, especially for
trials.

Adding next/previous links made this quicker and easier,
where as previously they could only navigate via the list
of hearings, which was inefficient and frustrating for
longer cases.

#### How
Unfortunately the usual pagination gems (will_paginate
and kaminari) are not going to work because:

a) we are consuming an API, not Active record models

b) although we are using json_api_client gem, that
   follows the jsonapi specification and is [supposed
   to be compatible with kaminari and will_paginate](https://github.com/JsonApiClient/json_api_client#library-compatibility)
   the Court data adaptor api itself does not implement the [jsonapi pagination specification](https://jsonapi.org/format/#fetching-pagination).

c) The nature of data coming from common platform and then from the adaptor
   does not lend itself to "out of the box" pagination solutions and
   a custom solution is likely to be more flexible while we progress our
   understanding and requirements.

--------

#### TODO (wip)

 - [x] speed improvement options: attempt to pass a cached prosecution case;
       pass a PageItem set; use caching of api responses
       - have now modified the hearing controller to just query the prosecution case endpoint and retrieve the hearing
       from here, rather than both the pc endpoint and hearing endpoint. I have also spiked caching locally which worked
       but this needs a separate ticket as infrastructure will be needed
 - [x] write issue on CDA for hearing_day filter on hearing endpoint
 - [x] write issue on CDA for jsonapi pagination
 - [x] write issue on CDA for jsonapi sorting

### DIFF
--- a/app/helpers/hearing_paginator.rb
+++ b/app/helpers/hearing_paginator.rb
@@ -17,8 +17,6 @@ class HearingPaginator
     @current_page = page.to_i
   end
 
-  attr_writer :current_page
-
   def current_page
     @current_page ||= 0
   end
--- a/app/helpers/hearing_paginator.rb
+++ b/app/helpers/hearing_paginator.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+# Creates a chronologically sorted set of "items"
+# with next and previous navigation helpers.
+#
+# Next and previous correlate to navigation over all
+# a cases' hearing's hearing days.
+#
+class HearingPaginator
+  include ActionView::Helpers
+  include Rails.application.routes.url_helpers
+
+  PageItem = Struct.new(:id, :hearing_date)
+
+  # rubocop:disable Rails/HelperInstanceVariable
+  def initialize(prosecution_case, page: 0)
+    @prosecution_case = prosecution_case
+    @current_page = page.to_i
+  end
+  # rubocop:enable Rails/HelperInstanceVariable
+
+  def items
+    sorted_hearing_page_items
+  end
+
+  def current_item
+    items[current_page || first_page]
+  end
+
+  attr_writer :current_page
+
+  def current_page
+    @current_page ||= 0
+  end
+
+  def first_page
+    0
+  end
+
+  def last_page
+    [items.size - 1, 0].max
+  end
+
+  def first_page?
+    current_page == first_page
+  end
+
+  def last_page?
+    current_page == last_page
+  end
+
+  def next_page_link
+    link_to(t('hearings.show.pagination.next_page'),
+            hearing_path(id: next_item.id,
+                         urn: prosecution_case.prosecution_case_reference,
+                         page: next_page),
+            class: 'govuk-link app-pagination-next')
+  end
+
+  def previous_page_link
+    link_to(t('hearings.show.pagination.previous_page'),
+            hearing_path(id: previous_item.id,
+                         urn: prosecution_case.prosecution_case_reference,
+                         page: previous_page),
+            class: 'govuk-link app-pagination-previous')
+  end
+
+  private
+
+  def next_item
+    items[next_page]
+  end
+
+  def next_page
+    [current_page + 1, last_page].min
+  end
+
+  def previous_item
+    items[previous_page]
+  end
+
+  def previous_page
+    [current_page - 1, first_page].max
+  end
+
+  def sorted_hearing_page_items
+    hearing_page_items.sort_by(&:hearing_date)
+  end
+
+  attr_reader :prosecution_case
+
+  def hearing_page_items
+    prosecution_case.hearings.each_with_object([]) do |hearing, result|
+      hearing.hearing_days.map do |day|
+        result << PageItem.new(hearing.id, day.to_datetime)
+      end
+    end
+  end
+end
--- a/app/helpers/hearing_paginator.rb
+++ b/app/helpers/hearing_paginator.rb
@@ -50,19 +50,21 @@ class HearingPaginator
   end
 
   def next_page_link
+    # TODO: add <span class="govuk-visually-hidden"> set of pages</span> within hyperlink ?!
     link_to(t('hearings.show.pagination.next_page'),
             hearing_path(id: next_item.id,
                          urn: prosecution_case.prosecution_case_reference,
                          page: next_page),
-            class: 'govuk-link app-pagination-next')
+            class: 'moj-pagination__link')
   end
 
   def previous_page_link
+    # TODO: add <span class="govuk-visually-hidden"> set of pages</span> within hyperlink ?!
     link_to(t('hearings.show.pagination.previous_page'),
             hearing_path(id: previous_item.id,
                          urn: prosecution_case.prosecution_case_reference,
                          page: previous_page),
-            class: 'govuk-link app-pagination-previous')
+            class: 'moj-pagination__link')
   end
 
   private
--- a/app/helpers/hearing_paginator.rb
+++ b/app/helpers/hearing_paginator.rb
@@ -12,19 +12,9 @@ class HearingPaginator
 
   PageItem = Struct.new(:id, :hearing_date)
 
-  # rubocop:disable Rails/HelperInstanceVariable
   def initialize(prosecution_case, page: 0)
     @prosecution_case = prosecution_case
     @current_page = page.to_i
-  end
-  # rubocop:enable Rails/HelperInstanceVariable
-
-  def items
-    sorted_hearing_page_items
-  end
-
-  def current_item
-    items[current_page || first_page]
   end
 
   attr_writer :current_page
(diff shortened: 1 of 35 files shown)


app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,99 @@
...
+  def next_page_link
+    link_to('Next hearing day',
+            hearing_path(id: next_item.id,
+                         urn: prosecution_case.prosecution_case_reference,
+                         page: next_page),
+            class: 'govuk-link app-pagination-next')
+  end
+
+  def previous_page_link
+    link_to('Previous hearing day',
```
jsugarman, line 61: move to translations

app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,99 @@
...
+  def first_page?
+    current_page == first_page
+  end
+
+  def last_page?
+    current_page == last_page
+  end
+
+  def next_page_link
+    link_to('Next hearing day',
```
jsugarman, line 53: move to translations

spec/requests/hearings_spec.rb
```diff
@@ -19,15 +26,15 @@
...
+      get "/hearings/#{hearing_id_from_fixture}?page=0&urn=#{case_reference}"
     end
 
     it { expect(response).to render_template('hearings/show') }
   end
 
   context 'when not authenticated' do
     before do
-      get "/hearings/#{hearing_id_from_fixture}&urn=#{case_reference}"
+      get "/hearings/#{hearing_id_from_fixture}&page=0urn=#{case_reference}"
```
jsugarman, line 37:
```suggestion
      get "/hearings/#{hearing_id_from_fixture}?page=0&urn=#{case_reference}"
```

spec/features/hearing_pagination_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      scenario 'user can see Next page navigation only' do
+        expect(page).not_to have_link('Previous hearing day')
+        expect(page).to have_link('Next hearing day')
+      end
+    end
+
+    context 'with a "middle" hearing\'s day displayed' do
+      before { click_link('26/10/2019') }
+
+      scenario 'user can see Next page navigation only' do
```
jsugarman, line 46:
```suggestion
      scenario 'user can see Next and previous page navigation link' do
```

app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,99 @@
...
+  def previous_item
+    items[previous_page]
+  end
+
+  def previous_page
+    [current_page - 1, first_page].max
+  end
+
+  def sorted_hearing_page_items
+    hearing_page_items.sort_by(&:hearing_date)
```
kmahern, line 87: Having pulled this branch and tested, the sorted_hearing_page_items order here does not necessarily match the order of the hearing dates in the Hearings table, if the dates of the Hearings overlap (possibly an edge case???). This means that when paging through hearing days the order doesn't necessarily match the order in the Hearings table (which I think DF said was a requirement) but also (more seriously) means that when you click on a date in the Hearings table you could get taken to the wrong day. As mentioned above I think this is probably only an issue when the hearing days for 2 hearings overlap.

spec/helpers/hearing_paginator_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+    it {
+      is_expected.to have_link('Previous hearing day',
+                               href: %r{/hearings/#{hearing2.id}\?page=2&urn=ACASEURN})
+    }
+
+    it {
+      is_expected.to have_link('Previous hearing day',
+                               class: 'govuk-link app-pagination-previous')
+    }
+  end
```
kmahern, line 256: Should this spec be extended to include tests for describe '#hearing_page_items' and describe '#sorted_hearing_page_items'.

jsugarman: Those are private methods and therefore should not be tested directly.

To [quote others](https://anthonysciamanna.com/2016/02/14/should-private-methods-be-tested.html)
>> The short answer is that you shouldn’t test private methods directly, but only their effects on the public methods that call them. 

They are tested via the public `#items` method, which as you can see calls `#sorted_hearing_page_items`, which in turn calls `#hearing_page_item`

app/helpers/hearing_paginator.rb
jsugarman: I will need to look at this. I assume you have duplicated the problem @kmahern?! I am not sure how edge case it is. It could be caused by differences between `#hearings_with_day_by_datetime` and `#sorted_hearing_page_items` ordering.

app/webpack/stylesheets/application.scss
```diff
@@ -10,6 +10,7 @@ $govuk-font-url-function: frontend-font-url;
 $govuk-image-url-function: frontend-image-url;
 @import "~govuk-frontend/govuk/all";
 @import "core/typography";
+@import "@ministryofjustice/frontend/moj/all";
```
naseberry, line 13: Import moj frontend before typography.
The import order is important should we wish to override styles coming from either govuk or moj design systems.

app/helpers/hearing_paginator.rb
jsugarman: @kmahern I have [pushed this commit](https://github.com/ministryofjustice/laa-court-data-ui/pull/475/commits/a4fa57807ecdaa7b73fb348ee863cd15cead5fb1)  that should duplicate the ordering between
the case details "hearing summaries" table and the hearing details page pagination ordering. 

This should deal with the difference you encountered so it would be good if can you check this again.

spec/helpers/hearing_paginator_spec.rb
kmahern: 👍

app/helpers/hearing_paginator.rb
kmahern: 👍 Have retested and lgtm.

kmahern (approved):
lgtm 👍

jrmhaig (commented):
I've added some comments to the `HearingPaginator` class but nothing major and I'd not be opposed to merging.

app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,95 @@
...
+  PageItem = Struct.new(:id, :hearing_date)
+
+  # rubocop:disable Rails/HelperInstanceVariable
+  def initialize(prosecution_case, options = {})
+    @prosecution_case = prosecution_case
+    @current_page = options.fetch(:page, nil).to_i
+  end
+  # rubocop:enable Rails/HelperInstanceVariable
+
+  def items
```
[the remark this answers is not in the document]
jsugarman: done for this one

app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,95 @@
...
+# Next and previous correlate to navigation over all
+# a cases' hearing's hearing days.
+#
+class HearingPaginator
+  include ActionView::Helpers
+  include Rails.application.routes.url_helpers
+
+  PageItem = Struct.new(:id, :hearing_date)
+
+  # rubocop:disable Rails/HelperInstanceVariable
```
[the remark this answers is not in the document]
jsugarman: somewhat bizarrely memoization is explicitly excluded by this cop as a result, apparently of this accepted issue: https://github.com/rubocop-hq/rubocop-rails/issues/311. I was thinking to raise my own issue related to using a class instance within a helper.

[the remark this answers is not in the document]
jsugarman: ...and in case you are thinking of the cheeky ;) workaround below
```
@current_page ||= page.to_i
```

it raises another cop violation
```
Lint/DisjunctiveAssignmentInConstructor: Unnecessary disjunctive assignment. Use plain assignment.
```

app/helpers/hearing_paginator.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def items
+    hearing_items_by_datetime
+  end
+
+  def current_item
+    items[current_page || first_page]
+  end
+
+  attr_writer :current_page
```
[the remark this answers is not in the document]
jsugarman: > Move attr_writer to the top of the class.

I kept this together with the attribute reader version for clarity. In terms of it being at the top what is the advantage? and what would the top be? above the `include` statements and Struct definition, which does not seem right to me, or below.

> Also, if it is not required for the public API of the class consider removing it and using @current_page instead.

true, albeit that the public API is somewhat in flux at the moment (there are other methods which do not need to be public too). current_page is used in the tests for convenience at the moment so many would have to be rewritten too - I will look at cleaning this up.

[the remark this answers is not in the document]
jsugarman: have removed the `current_page` `attr_writer` now and rewritten the tests.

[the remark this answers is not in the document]
jrmhaig: > In terms of it being at the top what is the advantage?

Only because I thought that it is generally good practice to have them listed at the top of a class. (Did I imagine that?)

> > Also, if it is not required for the public API of the class consider removing it and using @current_page instead.
> 
> true, albeit that the public API is somewhat in flux at the moment

Having things public when they don't need to be could make it trickier to change later.

> current_page is used in the tests for convenience at the moment so many would have to be rewritten too

I did see that and I think it would be better to have the class created as it would actually happen (by passing the correct parameters) rather than modifying internal values. If you modify internal values then there could come a time where we are testing situations that cannot occur due to changes in the interface. (I believe you have already updated the tests along these lines)

app/helpers/hearing_paginator.rb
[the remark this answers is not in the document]
jrmhaig: You can have `@current_page = page.to_i` in the initializer as it it known to be `nil` at that point. In the `current_page` method below (now lines 20-22) the check (I think) is not necessary as it has already been set.

app/helpers/hearing_paginator.rb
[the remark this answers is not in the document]
jsugarman: > have removed the `current_page` `attr_writer` now and rewritten the tests.

jrmhaig (approved):

PR #485, opened 2021-02-23
Cbo 1699 hearings sort order

kmahern:
#### What
Allow caseworkers to be able to sort the columns of the hearings table in ascending and descending order. Date is the most important attribute to sort by.

#### Ticket
https://dsdmoj.atlassian.net/browse/CBO-1416

#### Why
here are some scenarios where caseworkers don't need to review all hearings, but instead are only interested in certain specific hearings. Appeals are one common example of this. A descending sort on date would allow them to quickly see the last hearings, even with a long list of hearings.

#### How
Introduced new sort_order paramater; with options date_asc, date_desc, type_asc, type_desc, provider_asc and provider desc. The sort_order of the hearings table must be reflected in pagination, so that when a user navigates between hearings the order reflects the sort_order of the hearings table.


jsugarman (commented):
will need rebasing on mine to clear out unrelated changes in it, or wait and rebase on master soon. 

I was think along similar lines but using some mechanism of passing a sort "method" and sort "order" to the paginator.

e.g.

```
HearingPaginator.new(prosecution_case, page: x, sort_method: :hearing_by_datetime, sort_order: :asc)
```

and in controller:

```
HearingPaginator.new(prosecution_case, page: params[:page], sort_method: params[:sort_method],  sort_order: params[sort_order])

**Comment by jsugarman** [comment]
@kmahern also, if you make it a PR against my branch rather than master that will clear out my changes and just display your commits - needs rebasing though as mentioned

File: app/decorators/prosecution_case_decorator.rb [hunk 1]
```diff
@@ -23,4 +21,31 @@ def hearings_with_day_by_datetime
   def cracked?
     hearings.map { |h| h.cracked_ineffective_trial&.cracked? }.any?
   end
+
+  private
+
+  def sorted_hearings
```
jsugarman, line 27: Have been reading about refactoring and this method is starting to smell of needing an "extract class" using inheritance or composition scenario.

[very roughly](https://refactoring.guru/extract-subclass) but if you have martin fowlers "Refactoring" or Sandi Metz's "99 Bottles of oop" they have example and explanations

kmahern: @jsugarman I have attempted to create an extract class for this method by creating a TableSorter module containing a HearingsSorter class, with subclasses HearingsProviderSorter and HearingsTypeSorter. Is this the kind of thing you had in mind and if so is the decorators folder the right location (or should they be under /lib for example)?

app/decorators/table_sorters/hearings_sorter.rb
```diff
@@ -0,0 +1,29 @@
...
+        hearing.hearing_days.map(&:to_datetime).sort.reverse
+      else
+        hearing.hearing_days.map(&:to_datetime).sort
+      end
+    end
+
+    private
+
+    def order_by_asc_or_desc(sorted_hearings)
+      sorted_hearings.reverse! if @sort_order&.include? 'desc'
```
jrmhaig, line 25: 1. This will modify `sorted_hearings`. Would it work to do this?

```suggestion
      return sorted_hearings.reverse if @sort_order&.include? 'desc'
```

2. Is there a reason not to have `@sort_order == 'desc'`?
3. I'd suggest changing the argument so that it doesn't mask the `sorted_hearings` method. I just found it a little confusing.

jrmhaig: Regarding (2), I have just seen that the order can be `date_desc`.

app/decorators/table_sorters/hearings_sorter.rb
```diff
@@ -0,0 +1,29 @@
...
+      order_by_asc_or_desc(@hearings.sort_by { |h| h.hearing_days.map(&:to_datetime) })
+    end
+
+    def sorted_hearing(hearing)
+      if @sort_order.eql?('date_desc')
+        hearing.hearing_days.map(&:to_datetime).sort.reverse
+      else
+        hearing.hearing_days.map(&:to_datetime).sort
+      end
+    end
```
jrmhaig, line 20: If I understand correctly, the reversed option here will only happen when sorting by date, so it will never occur in either of the child classes `HearingsProviderSorter` or `HearingsTypeSorter`. It might be worth having third class `HearingsDateSorter` (used on line 34 of `app/decorators/prosecution_case_decorator.rb`) something like:

```ruby
module TableSorter
  class HearingsSorter

    ...

    def sorter_hearing(hearing)
      hearing.hearing_day.map(&:to_datetime).sort
    end

    ...

  end
end

module TableSorter
  class HearingsDateSorter << HearingsSorter
    def sorter_hearing(hearing)
      order_by_asc_or_desc(super)
    end
  end
end
```

kmahern: @jrmhaig I have created  a new HearingsDateSorter class which subclasses HearingsSorter and implements sorted_hearing as you suggest.

app/decorators/table_sorters/hearings_sorter.rb
kmahern: @jrmhaig, thanks for the suggestions, I have made the change you suggested in 1. and renamed the argument for order_by_asc_or_desc to arr.

app/helpers/application_helper.rb
```diff
@@ -36,4 +43,19 @@ def decorator_instance(object, decorator_class = nil)
     decorator_class ||= "#{object.class.to_s.demodulize}Decorator".constantize
     decorator_class.new(object, self)
   end
+
+  def column_sort_icon
```
jsugarman, line 47: ideally part of a decorator

app/helpers/application_helper.rb
```diff
@@ -36,4 +43,19 @@ def decorator_instance(object, decorator_class = nil)
...
+    sort_direction == 'asc' ? "\u25B2" : "\u25BC"
+  end
+
+  def column_title(column)
```
jsugarman, line 51: ideally part of a decorator

app/decorators/table_sorters/hearings_sorter.rb
```diff
@@ -0,0 +1,21 @@
...
+    def initialize(hearings, sort_order)
+      @hearings = hearings
+      @sort_order = sort_order
+    end
+
+    def sorted_hearing(hearing)
+      hearing.hearing_days.map(&:to_datetime).sort
+    end
+
+    private
```
jsugarman, line 14: should probably be protected rather than private

app/decorators/prosecution_case_decorator.rb
```diff
@@ -23,4 +21,17 @@ def hearings_with_day_by_datetime
   def cracked?
     hearings.map { |h| h.cracked_ineffective_trial&.cracked? }.any?
   end
+
+  private
+
+  def sorter
```
jsugarman, line 27: i was hoping this conditional would be got rid of by the refactor to use classes

jsugarman (commented):
are we not conflating two different concepts, :field and :order, or `:column` and `:direction`

Should a sorter not take two args and return. seems confused/ing

app/decorators/prosecution_case_decorator.rb
jsugarman: This is fundamentally a factory and should be a class method belonging to another class or module

jsugarman: e.g.
```
TableSorter.for(sort_order)
```

jsugarman:
```
module TableSorters
  def self.for(field = nil)
    case field
    when /^type/
      HearingsTypeSorter.new(hearings, field)
    when /^provider/
      HearingsProviderSorter.new(hearings, field)
    else
      HearingsDateSorter.new(hearings, field)
    end
  end
end
```

app/helpers/application_helper.rb
kmahern: 👍 I have moved column_title to the prosecution_case_decorator

app/decorators/table_sorters/hearings_sorter.rb
kmahern: 👍 I have changed this to protected

app/decorators/prosecution_case_decorator.rb
kmahern: 👍 As suggested, I have used the factory pattern to move the case statement to HearingsSorter

kmahern:
@jsugarman, I have removed all references to sort_order and replaced them with column and direction.

app/decorators/prosecution_case_decorator.rb
```diff
@@ -23,4 +21,25 @@ def hearings_with_day_by_datetime
...
+  def column_title(column)
+    case column
+    when 'type'
+      t('search.result.hearing.hearing_type')
+    when 'provider'
+      t('search.result.hearing.providers')
+    else
+      t('search.result.hearing.hearing_day')
+    end
+  end
```
jsugarman, line 38: **not a blocker**

A more OO alternative
```
def column_title
  Hash.new(t('search.result.hearing.hearing_day')).merge(
    type: t('search.result.hearing.hearing_type'),
    provider: t('search.result.hearing.providers')
  )[column]
end
```
Column is attribute on the object anyway I believe, so does it need passing to it. Although you may, and probably should, handle nil scenario for the attribute in here, rather than in the controller.

app/decorators/prosecution_case_decorator.rb
```diff
@@ -1,18 +1,16 @@
 # frozen_string_literal: true
 
 class ProsecutionCaseDecorator < BaseDecorator
+  attr_accessor :column, :direction
```
jsugarman, line 4: **not a blocker**
the name `column` in the context of this class does not clarify that it is actually a hearing column that is being referred to and the direction is the direction for hearing display.

I am not wanting to block this further but are we missing a class here - like HearingsDecorator that should know about decorating a list of hearings

config/locales/en/views/hearings.yml
```diff
@@ -19,7 +19,7 @@ en:
           no_hearing_details: No hearing details available
       hearing_type: Hearing type
       pagination:
-        next_page: Next hearing day
-        previous_page: Previous hearing day
+        next_page: Next
+        previous_page: Previous
```
jsugarman, line 23: Did dan request this change? He had explicitly requested "Next/Previous hearing day" before

spec/decorators/prosecution_case_decorator_spec.rb
```diff
@@ -17,6 +19,10 @@
     let(:object) { prosecution_case }
   end
 
+  it { is_expected.to respond_to(:column) }
+
+  it { is_expected.to respond_to(:direction) }
```
jsugarman, line 24:
```
it { is_expected.to respond_to(:colum, :column=, :direction, :direction=) }
```

spec/decorators/prosecution_case_decorator_spec.rb
```diff
@@ -43,50 +49,132 @@
...
+
+      let(:expected_days) do
+        ['2021-01-20T16:00:00.000Z'.to_datetime,
+         '2021-01-18T11:00:00.000Z'.to_datetime,
+         '2021-01-19T10:45:00.000Z'.to_datetime,
+         '2021-01-20T10:45:00.000Z'.to_datetime]
+      end
+
+      it { expect(call.map(&:day)).to eql(expected_days) }
+    end
```
jsugarman, line 123: this "says" it is testing `type` sorting but is testing based on date/day. would be confusing

spec/decorators/prosecution_case_decorator_spec.rb
```diff
@@ -43,50 +49,132 @@
...
+
+      let(:expected_days) do
+        ['2021-01-19T10:45:00.000Z'.to_datetime,
+         '2021-01-20T10:45:00.000Z'.to_datetime,
+         '2021-01-18T11:00:00.000Z'.to_datetime,
+         '2021-01-20T16:00:00.000Z'.to_datetime]
+      end
+
+      it { expect(call.map(&:day)).to eql(expected_days) }
+    end
```
jsugarman, line 139: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/485/files#r605680744)

spec/decorators/prosecution_case_decorator_spec.rb
```diff
@@ -43,50 +49,132 @@
...
+
+      let(:expected_days) do
+        ['2021-01-18T11:00:00.000Z'.to_datetime,
+         '2021-01-20T16:00:00.000Z'.to_datetime,
+         '2021-01-19T10:45:00.000Z'.to_datetime,
+         '2021-01-20T10:45:00.000Z'.to_datetime]
+      end
+
+      it { expect(call.map(&:day)).to eql(expected_days) }
+    end
```
jsugarman, line 158: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/485/files#r605680744) but for provider

spec/decorators/prosecution_case_decorator_spec.rb
```diff
@@ -43,50 +49,132 @@
...
+
+      let(:expected_days) do
+        ['2021-01-19T10:45:00.000Z'.to_datetime,
+         '2021-01-20T10:45:00.000Z'.to_datetime,
+         '2021-01-20T16:00:00.000Z'.to_datetime,
+         '2021-01-18T11:00:00.000Z'.to_datetime]
+      end
+
+      it { expect(call.map(&:day)).to eql(expected_days) }
+    end
```
jsugarman, line 177: [ditto](https://github.com/ministryofjustice/laa-court-data-ui/pull/485/files#r605680744) but for provider

spec/decorators/table_sorters/hearings_sorter_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    let(:expected_result) do
+      ['2021-01-19T10:45:00.000Z'.to_datetime, '2021-01-20T10:45:00.000Z'.to_datetime]
+    end
+
+    it {
+      is_expected.to eql(expected_result)
+    }
+  end
+
+  describe '#self.for' do
```
jsugarman, line 40:
```suggestion
  describe '.for' do
```

spec/features/hearing_sorting_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+        expect(rows[5]).to have_link('31/10/2019', href: hearing_page_url(5, 'type', 'desc'),
+                                                   class: 'govuk-link')
+        expect(rows[6]).to have_link('26/10/2019', href: hearing_page_url(6, 'type', 'desc'),
+                                                   class: 'govuk-link')
+      end
+    end
+
+    scenario 'user can sort by type asc' do
+      click_link('Hearing type')
+      click_link('Hearing type')
```
jsugarman, line 81: is this double click deliberate

jsugarman: looks like it is 👍

app/decorators/prosecution_case_decorator.rb
```diff
@@ -1,18 +1,16 @@
...
-    hearings.sort_by { |h| h.hearing_days.map(&:to_datetime) }
-  end
-
-  def hearings_with_day_by_datetime
+  def sorted_hearings_with_day
     Enumerator.new do |enum|
-      hearings_by_datetime.each do |hearing|
-        hearing.hearing_days.map(&:to_datetime).sort.each do |day|
+      sorter.sorted_hearings.each do |hearing|
+        sorter.sorted_hearing(hearing).each do |day|
```
jsugarman, line 13: although more verbose `sorted_hearing_days` is more declarative

jsugarman:
No blockers per-se on the functional code side of things, although i think the controller should probably not know about defaults for hearings sorting since we have classes that can handle and isolate that.

The naming of the attributes could also be misleading in the context of a prosecution_case - `prosecution_case.column` and `prosecution_case.direction` do not really state there intention (i.e. hearing sort column and  direction)

Couple of factory related suggestions, but not a blocker

On the specs I think there is a lot of repetition that might be handleable by shared examples, plus my other concerns regarding what some tests are basing success on - [example](https://github.com/ministryofjustice/laa-court-data-ui/pull/485/files#discussion_r605682032)

app/decorators/prosecution_case_decorator.rb
kmahern: @jsugarman, followed the suggestion to use a hash rather than a case statement here. I am also now handling the defaults for hearings_sort_column and hearings_sort_direction here rather than in the controller.

app/decorators/prosecution_case_decorator.rb
kmahern: I have amended column and direction to hearings_sort_column and hearings_sort_direction.

spec/decorators/prosecution_case_decorator_spec.rb
kmahern: 👍 I have made the suggested change.

spec/decorators/prosecution_case_decorator_spec.rb
kmahern: 👍 I have amended this to test expect(call.map(&:hearing_type)) to attempt to make this clearer

spec/decorators/prosecution_case_decorator_spec.rb
kmahern: 👍 I have amended this to test expect(call.map(&:hearing_type)) to attempt to make this clearer

spec/decorators/prosecution_case_decorator_spec.rb
kmahern: 👍 I have amended this to test expect(call.map(&:provider_list)) to attempt to make this clearer

spec/decorators/prosecution_case_decorator_spec.rb
kmahern: 👍 I have amended this to test expect(call.map(&:provider_list)) to attempt to make this clearer

spec/decorators/table_sorters/hearings_sorter_spec.rb
kmahern: 👍 I have made the suggested change.

app/decorators/prosecution_case_decorator.rb
kmahern: 👍 I have amended sorted_hearing to sorted_hearing_days

kmahern:
@jsugarman I think I have made the changes you suggest and have created shared_examples_for_hearings_sorters.rb to attempt to reduce repetition.

config/locales/en/views/hearings.yml
kmahern: I have just checked with Dan and he agreed with this change as you would now be able to page through the days in different (non-chronological) order.

app/decorators/table_sorters/hearings_sorter.rb
```diff
@@ -0,0 +1,29 @@
...
+      @hearings = hearings
+      @column = column
+      @direction = direction
+    end
+
+    def self.for(hearings, column, direction)
+      Hash.new(TableSorters::HearingsDateSorter.new(hearings, column, direction)).merge(
+        type: TableSorters::HearingsTypeSorter.new(hearings, column, direction),
+        provider: TableSorters::HearingsProviderSorter.new(hearings, column, direction)
+      )[column.to_sym]
```
jsugarman, line 15: You  don't need to repeat the `.new(hearings, column, direction)`

```suggestion
      Hash.new(TableSorters::HearingsDateSorter).merge(
        type: TableSorters::HearingsTypeSorter,
        provider: TableSorters::HearingsProviderSorter
      )[column.to_sym].new(hearings, column, direction)
```

spec/decorators/table_sorters/hearings_date_sorter_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context 'when column is date and sort direction is asc' do
+      let(:column) { 'date' }
+      let(:direction) { 'asc' }
+      let(:expected_result) do
+        ['2021-01-19T10:45:00.000Z'.to_datetime, '2021-01-20T10:45:00.000Z'.to_datetime]
+      end
+
+      it {
+        is_expected.to eql(expected_result)
+      }
```
jsugarman, line 48: **nit**: one liner

spec/decorators/table_sorters/hearings_date_sorter_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context 'when column is date and sort direction is desc' do
+      let(:column) { 'date' }
+      let(:direction) { 'desc' }
+      let(:expected_result) do
+        ['2021-01-20T10:45:00.000Z'.to_datetime, '2021-01-19T10:45:00.000Z'.to_datetime]
+      end
+
+      it {
+        is_expected.to eql(expected_result)
+      }
```
jsugarman, line 60: **nit**: one liner

jsugarman: or do end with description for consistency with other similar tests

spec/decorators/table_sorters/hearings_date_sorter_spec.rb
jsugarman: or do end with description for consistency with other similar tests

spec/decorators/table_sorters/hearings_sorter_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+    end
+
+    context 'when column is type and direction is asc' do
+      let(:column) { 'type' }
+      let(:direction) { 'asc' }
+
+      it {
+        is_expected.to be_instance_of(TableSorters::HearingsTypeSorter)
+      }
```
jsugarman, line 42: **nit**: one liner

spec/decorators/table_sorters/hearings_sorter_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    context 'when column is date and direction is asc' do
+      let(:column) { 'date' }
+      let(:direction) { 'asc' }
+
+      it {
+        is_expected.to be_instance_of(TableSorters::HearingsDateSorter)
+      }
```
jsugarman, line 51: **nit**: one liner

jsugarman: unless rubocop complains

spec/decorators/table_sorters/hearings_sorter_spec.rb
jsugarman: unless rubocop complains

spec/decorators/table_sorters/hearings_sorter_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  describe '.for' do
+    subject { described_class.for(hearings, column, direction) }
+
+    context 'when column is provider and direction is asc' do
+      let(:column) { 'provider' }
+      let(:direction) { 'asc' }
+
+      it 'sorts hearing by hearing_days desc' do
+        is_expected.to be_instance_of(TableSorters::HearingsProviderSorter)
+      end
```
jsugarman, line 33: This description looks like it should be empty/consistent with examples elow

```suggestion
      it {
        is_expected.to be_instance_of(TableSorters::HearingsProviderSorter)
      }
```

spec/decorators/table_sorters/hearings_sorter_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    context 'when column is date and direction is asc' do
+      let(:column) { 'date' }
+      let(:direction) { 'asc' }
+
+      it {
+        is_expected.to be_instance_of(TableSorters::HearingsDateSorter)
+      }
+    end
+  end
```
jsugarman, line 53: should probably have an example to explicitly test the default, and arguably, do not need the above test for `'date'` as this equates to the default.

```
    context 'when column and direction not provided' do
      let(:column) { nil }
      let(:direction) { nil }

      it {
        is_expected.to be_instance_of(TableSorters::HearingsDateSorter)
      }
    end
```

jsugarman: also the `direction` has no bearing on the the outcome so could be passed as `nil` for all these tests and would be clearer in that context.

app/decorators/prosecution_case_decorator.rb
```diff
@@ -23,4 +21,29 @@ def hearings_with_day_by_datetime
...
+
+  def column_sort_icon
+    hearings_sort_direction == 'asc' ? "\u25B2" : "\u25BC"
+  end
+
+  def column_title(column)
+    Hash.new(t('search.result.hearing.hearing_day')).merge(
+      type: t('search.result.hearing.hearing_type'),
+      provider: t('search.result.hearing.providers')
+    )[column.to_sym]
```
jsugarman, line 33: **not a blocker** probably safer to a safe navigation operator here
```suggestion
    )[column&.to_sym]
```

app/decorators/table_sorters/hearings_sorter.rb
jsugarman: have added a commit to this affect

spec/decorators/table_sorters/hearings_sorter_spec.rb
jsugarman: have applied a commit this this affect

spec/decorators/table_sorters/hearings_sorter_spec.rb
jsugarman: have add a commit this this affect

jsugarman (approved):
Couple of minor test related tweaks remain in the comments but no blockers. I think the naming of the sorters needs attention but am not going to block this further.

PR #515, opened 2021-03-23
[CBO-1756] partial name search

jsugarman:
#### What
Partial name search

#### Ticket
[CBO-1756](https://dsdmoj.atlassian.net/browse/CBO-1756)

#### Why
Enable case workers to search for defendants using
partial name matching.

#### How
Pass search term direct to adaptor and do not
filter response/results.

Since common platform will search
by name case-insensitively and apply
partial matching VCD and CDA do not
need to manipulate the search term
nor the results returned.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -35,7 +35,7 @@ class Search
 
   validates :term,
             presence: true,
-            format: { with: /\A[A-Za-z0-9\s]+\z/ },
+            format: { with: /\A[A-Za-z0-9\s\-']+\z/ },
             char_length: { minimum: 2 }
 
   validates :dob,
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -35,6 +35,7 @@ class Search
 
   validates :term,
             presence: true,
+            format: { with: /\A[A-Za-z0-9\s]+\z/ },
             char_length: { minimum: 2 }
 
   validates :dob,
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -33,7 +33,7 @@ class Search
     message: 'Filter "%{value}" is not recognized'
   }
 
-  validates :term, presence: true
+  validates :term, presence: true, visible_length: { minimum: 2 }
   validates :dob,
             presence: true,
             if: proc { |search| search.filter.eql?('defendant_name') }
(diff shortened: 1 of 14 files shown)


spec/lib/court_data_adaptor/query/defendant/by_name_spec.rb
```diff
@@ -26,53 +26,119 @@ def self.resource
...
+
+        let(:term) { 'trever' }
+
+        it 'returns defendant resources' do
+          expect(results).to all(be_instance_of(CourtDataAdaptor::Resource::Defendant))
+        end
+
+        it 'returns more than just exact match' do
+          expect(results.size).to be 1
+        end
```
jrmhaig, line 126: Is this 'more than just exact match'?

With this and the one above lines 90-92 I think it would be clearer if something identifying the correct result(s) is shown rather than a count.

jsugarman: Am getting rid of this check as is fairly meaningless, as name matching is handled in the spec at line 132

jsugarman: thats done now

jrmhaig (approved):
Looks OK.

jsugarman:
May need to limit chars to 2 or 3 for search to mitigate response timeouts seen in SIT

app/models/search.rb
```diff
@@ -28,12 +28,16 @@ def filters
...
-    message: 'Filter "%{value}" is not recognized'
-  }
+  validates :filter,
+            presence: true,
+            inclusion: { in: filters.map { |f| f.id.to_s },
+                         message: 'Filter "%{value}" is not recognized' }
+
+  validates :term,
+            presence: true,
+            format: { with: /\A[A-Za-z0-9\s]+\z/ },
```
kmahern, line 38: Is there a potential issue here with names containing characters (like O'Reilly?)

jsugarman: handle. have been talking to alan and testing. see last few commits - basically allow apostrophes but not hyphens (es handles these in unexpected ways and can timeout/error due to CDA issues.

kmahern (approved):
👍

PR #519, opened 2021-03-26
Bump mimemagic 0.3.9

jsugarman:
#### What
Bump mimemagic 0.3.9

#### Why
see https://www.theregister.com/2021/03/25/ruby_rails_code/

Previous versions to 0.3.6 yanked
plus 0.3.9 seems to just work.

We do not actually use it anyway, but is a rails
dependency that we are currently pulling in,
even though it should be possible not to by removing
`require 'rails'` from `config/application.rb`

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 brew 'postgresql', restart_service: true
+brew 'shared-mime-info'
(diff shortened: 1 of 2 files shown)


jsugarman:
Closing as superceded by https://github.com/ministryofjustice/laa-court-data-ui/pull/522

PR #525, opened 2021-04-01
[security] selectively bump is-svg from 3.0.0 to 4.3.1

naseberry:
CVE-2021-28092
high severity
Vulnerable versions: >= 2.1.0, < 4.2.2
Patched version: 4.2.2
The is-svg package 2.1.0 through 4.2.1 for Node.js uses a regular expression that is vulnerable to Regular Expression Denial of Service (ReDoS). If an attacker provides a malicious string, is-svg will get stuck processing the input for a very long time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "resolutions": {
     "dot-prop": "^5.2.0",
     "http-proxy": "^1.18.1",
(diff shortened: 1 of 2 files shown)


kmahern (approved):
👍 already implemented on CCCD

PR #534, opened 2021-04-14
Add CharLengthValidator spec

jsugarman:
#### What
Spec the validator that was added as
part of partial name search

#### Why
Was missed/not-added as part of partial name
search implementation

#### How
Add spec and a custom rspec matcher
to help with tests. Also rename the
previous have_activerecord_error custom
matcher to better reflects its use and
purpose and remove a redundant use of
it.


kmahern (approved):
👍

PR #552, opened 2021-05-04
Display CDA errors on failed link

kmahern:
#### Ticket

[board ticket description](link)

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


spec/lib/court_data_adaptor/resource/base_spec.rb
```diff
@@ -76,6 +76,21 @@ class CourtDataAdaptor::Resource::MockResource < described_class
...
+          stub_request(:get, mock_resource_endpoint)
+            .to_return(
+              status: 422,
+              body: { field: %w[error1 error2] }.to_json
+            )
+        end
+
+        it 'applies custom handler' do
+          expect { described_class.all }.to \
+            raise_error CourtDataAdaptor::Errors::BadRequest, 'Bad request'
```
laurentqro, line 91: how about `Unprocessable entity` as an error message instead of `Bad request` ?

spec/support/webmock.rb
```diff
@@ -106,21 +106,23 @@
     ).to_return(
       status: 400,
       headers: { 'Content-Type' => 'application/vnd.api+json' },
-      body: { defendant_id: ['is not a valid uuid'],
-              maat_reference: [
-                '1234567 has no common platform data created against Maat application.'
-              ] }.to_json
+      body: {
+        'error' =>
+          'Contract failed with: {:defendant_id=>[\"is not a valid uuid\"]},
```
laurentqro, line 111: CDA will return `422` for contract failures.

spec/lib/court_data_adaptor/resource/base_spec.rb
laurentqro: The class name `BadRequest` might also be misleading

spec/support/webmock.rb
kmahern: 👍 have corrected

spec/lib/court_data_adaptor/resource/base_spec.rb
kmahern: @laurentqro I have created error class UnprocessableEntity. I have also changed name of the class BadRequestHandler to ApiRequestHandler, as this class will now be responsible for handling both BadRequest (400) responses and UnprocessableEntitity (422) responses. As suggested, an UnprocessableEntity response will now have a message associated 'Unprocessable entity' (although currently we are not displaying this). What do you think?

laurentqro (approved):

PR #570, opened 2021-05-12
[AAC-62] Display event hearing notes

mabely:
#### What
Change View Court Data so that it can accept the free text field (notes) associated with various events sent from CP. Add small bump of space to indicate change between “hearing event” and “note”

#### Ticket
[Implement free text in VCD](https://dsdmoj.atlassian.net/browse/AAC-62)

#### Why
Court Clerks want to add free text so that they can send extra details related to hearing events. HMCTS has provided a new set of schemas that are intended to support the addition of free text to the hearing event log API. CDA has been modified to accommodate this new information & transmit this data from CPP to VCD.

New Information in relation to hearings will be added as a free text & this information is required for LAA billing caseworkers to evaluate bills effectively. 

#### How
- Display the notes field for each hearing event sent from Common Platform via CDA
- Ensure any UNICODE characters are supported
- Ensure free text does not create security vulnerabilities
- Ensure any HTML is escaped

### DIFF
--- a/spec/views/hearings/_hearing_events.html.haml_spec.rb
+++ b/spec/views/hearings/_hearing_events.html.haml_spec.rb
@@ -10,15 +10,18 @@ RSpec.describe 'hearings/_hearing_events.html.haml', type: :view do
   let(:hearing_day) { Date.parse('2021-01-17T10:30:00.000Z') }
 
   let(:hearing_event1) do
-    hearing_event_class.new(description: 'day 1 start', occurred_at: '2021-01-17T10:30:00.000Z')
+    hearing_event_class.new(description: 'day 1 start', note: '1 hour delay',
+                            occurred_at: '2021-01-17T10:30:00.000Z')
   end
 
   let(:hearing_event2) do
-    hearing_event_class.new(description: 'day 1 end', occurred_at: '2021-01-17T16:30:00.000Z')
+    hearing_event_class.new(description: 'day 1 end', note: 'ended late',
+                            occurred_at: '2021-01-17T16:30:00.000Z')
   end
 
   let(:hearing_event3) do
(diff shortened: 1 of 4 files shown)


spec/views/hearings/_hearing_events.html.haml_spec.rb
```diff
@@ -57,5 +60,49 @@
...
+        let(:hearing_events) { [hearing_event_with_html_notes] }
+
+        it 'displays any HTML raw' do
+          is_expected.to have_content('<b>1 hour delay</b>')
+        end
+      end
+
+      context 'with notes containing unicode characters' do
+        let(:hearing_event_with_unicode_notes) do
+          hearing_event_class.new(description: 'day 1 start', note: '!\"#£%&()*,-./',
```
MalcolmVonMoJ, line 85: Is this a check for unicode characters?  
Can we include some more characters here?  
`Æ ½ ŵ €`
The ŵ is probably the most important of these 4.

MalcolmVonMoJ: And `"" '' –` - the common MS Word Autocorrect characters.

mabely: Just added the 4 extra characters you suggested, the MS Word characters you mentioned are already included in the test case.

MalcolmVonMoJ (approved):
Looks good to me 👍

kmahern:
We could add a test to https://github.com/ministryofjustice/laa-court-data-ui/blob/master/spec/lib/court_data_adaptor/resource/hearing_event_spec.rb something like 

`it { is_expected.to respond_to :note }`

mabely:
@kmahern thanks for pointing that spec out, I'd missed it but have added the test in now

kmahern (approved):
lgtm 👍

PR #574, opened 2021-05-14
Add fix for hearings view where hearing_events is not present in response 

mabely:
#### What
An error is thrown when viewing a hearing where the hearing_events is not present. Part 2 to [this PR](https://github.com/ministryofjustice/laa-court-data-ui/pull/573), which fixes the case view.

#### Why
Error is encountered where hearing_events is not present, preventing the page from loading.

#### How
Adding check for if hearing_events is present before trying to render it.

### DIFF
--- a/app/views/hearings/_hearing_events.html.haml
+++ b/app/views/hearings/_hearing_events.html.haml
@@ -6,9 +6,10 @@
       %th.govuk-table__header{ scope: 'col' }= t('hearings.show.events.time')
       %th.govuk-table__header{ scope: 'col' }= t('hearings.show.events.event')
   %tbody.govuk-table__body
-    - hearing.hearing_events_for_day(hearing_day).each do |event|
-      %tr.govuk-table__row
-        %td.govuk-table__cell
-          = event.occurred_at.to_datetime.strftime('%H:%M')
-        %td.govuk-table__cell
-          = event.description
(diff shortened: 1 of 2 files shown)


laurentqro (approved):

app/views/hearings/_hearing_events.html.haml
```diff
@@ -6,9 +6,10 @@
       %th.govuk-table__header{ scope: 'col' }= t('hearings.show.events.time')
       %th.govuk-table__header{ scope: 'col' }= t('hearings.show.events.event')
   %tbody.govuk-table__body
-    - hearing.hearing_events_for_day(hearing_day).each do |event|
-      %tr.govuk-table__row
-        %td.govuk-table__cell
-          = event.occurred_at.to_datetime.strftime('%H:%M')
-        %td.govuk-table__cell
-          = event.description
+    - if hearing.hearing_events.present?
```
laurentqro, line 9: another potential solution, render the partial only if there are any hearing events:

`= render(partial: 'hearing_events', locals: { hearing: @hearing, hearing_day: @hearing_day }) if hearing.events`

on `app/views/hearings/show.html.haml` at line 9

mabely: Thanks, I moved it over to the show view as that makes more sense

PR #580, opened 2021-05-18
[AAC-54] events free text

mabely:
#### What
Change View Court Data so that it can display free text field (hearing event notes) associated with various hearing events sent from CP. Hearing event notes will be max 3000 characters. Add small bump of space to indicate change between “hearing event description” and “note”.

A [PR was previously opened](https://github.com/ministryofjustice/laa-court-data-ui/pull/570#issue-643325845) with comments. Two additional commits were made in this PR with changes to:
- Ensure carriage return and new line characters `\n` `\r` are displayed as new line, etc.
- Limit width of table column width to enforce a line break in case of long unbreakable string inputs

#### Ticket
[Implement free text in VCD](https://dsdmoj.atlassian.net/browse/AAC-62)

#### Why
Court Clerks want to add free text so that they can send extra details related to hearing events. HMCTS has provided a new set of schemas that are intended to support the addition of free text to the hearing event log API. CDA has been modified to accommodate this new information & transmit this data from CPP to VCD.

New Information in relation to hearings will be added as a free text & this information is required for LAA billing caseworkers to evaluate bills effectively. 

#### How
- Display the notes field for each hearing event sent from Common Platform via CDA
- Ensure any UNICODE characters are supported
- Sanitize against any unsafe inputs
- Ensure any HTML is escaped
- Ensure carriage return and new line characters (\n \r) are displayed as new line, etc.
- Limit width of table column width to enforce a line break in case of long unbreakable string inputs

### DIFF
--- a/app/helpers/hearing_helper.rb
+++ b/app/helpers/hearing_helper.rb
@@ -8,4 +8,15 @@ module HearingHelper
   def earliest_day_for(hearing)
     hearing&.hearing_days&.map(&:to_datetime)&.sort&.first
   end
+
+  def display_crlf_and_sanitize(event_note)
+    note_with_converted_crlf = convert_crlf_to_html(event_note)
+    sanitize(note_with_converted_crlf, tags: %w[br p])
+  end
+
+  private
+
+  def convert_crlf_to_html(event_note)
+    simple_format(event_note, {}, wrapper_tag: 'span')
+  end
 end
--- a/app/helpers/hearing_helper.rb
+++ b/app/helpers/hearing_helper.rb
@@ -9,7 +9,7 @@ module HearingHelper
     hearing&.hearing_days&.map(&:to_datetime)&.sort&.first
   end
 
-  def display_crlf_and_sanitize(event_note)
+  def transform_and_sanitize(event_note)
     note_with_converted_crlf = convert_crlf_to_html(event_note)
     sanitize(note_with_converted_crlf, tags: %w[br p])
   end
--- a/spec/helpers/hearing_helper_spec.rb
+++ b/spec/helpers/hearing_helper_spec.rb
@@ -35,8 +35,8 @@ RSpec.describe HearingHelper, type: :helper do
     it { is_expected.to eql('2021-01-19T10:45:15.000Z'.to_datetime) }
   end
 
-  describe '#display_crlf_and_sanitize' do
-    subject { helper.display_crlf_and_sanitize(event_note) }
+  describe '#transform_and_sanitize' do
+    subject { helper.transform_and_sanitize(event_note) }
 
     context 'with notes containing unsafe and unpermitted html' do
       let(:event_note) { '<b>warning</b> <script>alert(123)</script>' }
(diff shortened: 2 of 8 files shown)


app/helpers/hearing_helper.rb
```diff
@@ -8,4 +8,15 @@ def paginator(...)
   def earliest_day_for(hearing)
     hearing&.hearing_days&.map(&:to_datetime)&.sort&.first
   end
+
+  def display_crlf_and_sanitize(event_note)
+    note_with_converted_crlf = convert_crlf_to_html(event_note)
+    sanitize(note_with_converted_crlf, tags: %w[br p])
```
jsugarman, line 14: do you need to whitelist `span` tag too? as you are using this instead of default `p` tag at line `20` below

jsugarman: according to the definition of [santize here](https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize)

spec/helpers/hearing_helper_spec.rb
```diff
@@ -34,4 +34,26 @@
 
     it { is_expected.to eql('2021-01-19T10:45:15.000Z'.to_datetime) }
   end
+
+  describe '#format_and_dislay_crlf' do
```
jsugarman, line 38: is this  the old name? it should match the subject anyway
```suggestion
describe '#display_crlf_and_sanitize' do
```

app/helpers/hearing_helper.rb
mabely: `simple_format` wraps `<p>` tags around the string by default, which would add extra spacing unnecessarily since it is whitelisted (to display `\r`) so I had to set it to something else that's not whitelisted to prevent it . I could also whitelist the `<span>` for it to render but since it makes no difference to the view, I chose not to clutter the view with unnecessary html

spec/helpers/hearing_helper_spec.rb
mabely: Thanks for catching that

jsugarman (approved):
👍 Looks ok to me without having viewed the output that results myself.

Couple of considerations:

1. naming wise you could use something more generic like `transform_and_sanitize` or just `transform`
2. if it gets more complex i would be tempted to write your own `NoteSanitizer` class
3. if `\n<br>` is rendered as two line feeds in html perhaps you should just convert it to `<br><br>`

app/helpers/hearing_helper.rb
mabely: Both `\n\n` or `\r\n\r\n` are considered a paragraph and are wrapped in `p` tags. If `p` is not whitelisted, then those will be sanitized out. Apologies I'm not doing a good job of explaining. 

Just for ref:
```
my_text = "Here is some basic text...\n...with a line break."

simple_format(my_text)
# => "<p>Here is some basic text...\n<br />...with a line break.</p>"

simple_format(my_text, {}, wrapper_tag: "div")
# => "<div>Here is some basic text...\n<br />...with a line break.</div>"

```
[simple_format documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html)

kmahern (approved):
@mabely this looks good to merge if Alan is happy with the testing 👍

PR #585, opened 2021-05-19
Test hearing_events template is not rendered if hearing_events is missing

mabely:
#### What
Prosecution case page ([raw_response section](https://github.com/ministryofjustice/laa-court-data-ui/pull/573)) and hearings page ([hearing_events](https://github.com/ministryofjustice/laa-court-data-ui/pull/574/files)) were not displaying in VCD on UAT, causing the case page to return an error page when the user tries to view a case. A fix was put in to check that hearing_events is present before trying to render but it was missing tests.  

#### Ticket

[Add test for hearing events fix](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=622&projectKey=AAC&selectedIssue=AAC-83)

#### Why
The fix was made urgently and so it is missing a unit test. 

#### How
Test that if hearing_events is not present, the relevant partial will not be rendered.


kmahern (approved):
👍

PR #590, opened 2021-05-25
Update local CDA setup documentation

mabely:
#### What
Update local CDA setup documentation

#### Ticket

[board ticket description](https://dsdmoj.atlassian.net/browse/AAC-91)

#### Why
Some Postgres setup errors were encountered when setting up CDA locally according to the README.

#### How
To avoid possible Postgres errors encountered when trying to run `rails console` to generate OAuth2 credentials, the readme now sets the CDA database URL and creates the database first which will hopefully prevent this for others.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -21,7 +21,8 @@ cd .../hmcts-common-platform-mock-api
 - create dummy data
 ```
 # create some data
-$ rails console
+rails mock:demodata:load
+rails console
 Pry> FactoryBot.create(:realistic_prosecution_case, defendants: FactoryBot.create_list(:realistic_defendant, 3))
 ```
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -40,7 +41,20 @@ git clone git@github.com:ministryofjustice/laa-court-data-adaptor.git
...
+DATABASE_URL=postgres://localhost/laa_court_data_adaptor_development
+INLINE_SIDEKIQ=true
+```
+
+- setup CDA database
+```
+rails db:create db:migrate db:seed
+```
+
+- generate OAuth2 `client_credentials` - for the UI (see end of section for postgres error)
```
kmahern, line 57: @mabely not sure what is meant by see end of section for postgres error?

mabely: @kmahern I've removed that, must have forgotten to remove by accident. I also added an alternative to setting up the db in the same commit so it aligns with CDA documentation, what I added didn't work for me but just so it doesn't diverge I included it - just FYI.

docs/development.md
```diff
@@ -21,7 +21,8 @@ cd .../hmcts-common-platform-mock-api
 - create dummy data
 ```
 # create some data
-$ rails console
+rails mock:demodata:load
```
kmahern, line 24: @mabely I think if we document to run rails` mock:demodata:load` we do not need ```
rails console
Pry> FactoryBot.create(:realistic_prosecution_case, defendants: FactoryBot.create_list(:realistic_defendant, 3))
``` (as I think these would do the same thing).

kmahern:
One small comment otherwise lgtm 👍

kmahern (commented):
👍

kmahern (approved):

docs/development.md
mabely: Removed it - thanks for spotting it!

PR #605, opened 2021-06-07
Modify alpha/beta banner to display environment for non-prod envs

kmahern:
#### What
We need a banner on the UAT environment

#### Ticket
https://dsdmoj.atlassian.net/browse/AAC-79

#### Why
Some of our users have logged into UAT and reported bugs because they are using test data. We still need caseworkers and other “non-technical” users to be testing new features/data that comes through from Common Platform. A banner would help avoid any users making this mistake and accidentally causing disruption.

#### How
Followed the suggestion on the ticket of modifying the Alpha/Beta banner to display the environment with a different colour (NB this is only for the dev, staging and uat environment, all other environments should be as is).

### DIFF
--- a/app/views/layouts/_phase_banner.html.haml
+++ b/app/views/layouts/_phase_banner.html.haml
@@ -1,10 +1,10 @@
 .govuk-phase-banner{ class: ENV['ENV'] }
   %p.govuk-phase-banner__content
     %strong.govuk-tag.govuk-phase-banner__content__tag
-      - if %w(uat staging dev).include? ENV['ENV']
-        = ENV['ENV']
-      - else
+      - if ENV['ENV'] == 'production'
         = t('generic.alpha')
+      - else
+        = ENV['ENV'] || 'local'
     %span.govuk-phase-banner__text
       = t('layouts.application.header.phase_banner_html', feedback: new_feedback_path)
 
--- a/app/views/layouts/_phase_banner.html.haml
+++ b/app/views/layouts/_phase_banner.html.haml
@@ -1,7 +1,10 @@
 .govuk-phase-banner
   %p.govuk-phase-banner__content
     %strong.govuk-tag.govuk-phase-banner__content__tag
-      = t('generic.alpha')
+      - if ENV['ENV'] == 'production'
+        = t('generic.alpha')
+      - else
+        = ENV['ENV'] || ENV['RAILS_ENV']
     %span.govuk-phase-banner__text
       = t('layouts.application.header.phase_banner_html', feedback: new_feedback_path)
 
--- a/app/views/layouts/_phase_banner.html.haml
+++ b/app/views/layouts/_phase_banner.html.haml
@@ -1,10 +1,10 @@
-.govuk-phase-banner
+.govuk-phase-banner{ class: ENV['ENV'] }
   %p.govuk-phase-banner__content
     %strong.govuk-tag.govuk-phase-banner__content__tag
-      - if ENV['ENV'] == 'production'
-        = t('generic.alpha')
+      - if %w(uat staging dev).include? ENV['ENV']
+        = ENV['ENV']
       - else
-        = ENV['ENV'] || ENV['RAILS_ENV']
+        = t('generic.alpha')
     %span.govuk-phase-banner__text
       = t('layouts.application.header.phase_banner_html', feedback: new_feedback_path)
 
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -28,7 +28,7 @@
 
     = render partial: 'layouts/cookie_banner'
 
-    %header.govuk-header{"data-module" => "govuk-header", role: "banner"}
+    %header.govuk-header{"data-module" => "govuk-header", role: "banner", class: ENV['ENV'] }
       .govuk-header__container.govuk-width-container
         .govuk-header__logo
           = render 'layouts/govuk_logo'
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -20,7 +20,7 @@
     %link{href: asset_pack_path("media/images/govuk-apple-touch-icon.png"), rel: "apple-touch-icon"}/
     = tag :meta, property: 'og:image', content: asset_pack_path('media/images/govuk-opengraph-image.png')
     = render partial: 'layouts/google_analytics'
-  %body.govuk-template__body
+  %body.govuk-template__body{ class: app_environment }
     :javascript
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     %a.govuk-skip-link{href: "#main-content"}
@@ -28,7 +28,7 @@
 
     = render partial: 'layouts/cookie_banner'
 
-    %header.govuk-header{"data-module" => "govuk-header", role: "banner", class: ENV['ENV'] }
+    %header.govuk-header{"data-module" => "govuk-header", role: "banner" }
       .govuk-header__container.govuk-width-container
         .govuk-header__logo
           = render 'layouts/govuk_logo'
--- a/app/webpack/stylesheets/components/_phase_banner.scss
+++ b/app/webpack/stylesheets/components/_phase_banner.scss
@@ -1,8 +1,9 @@
 $app-uat-colour: govuk-colour('orange');
 $app-staging-colour: govuk-colour('turquoise');
 $app-dev-colour: govuk-colour('pink');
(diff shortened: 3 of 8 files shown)


app/webpack/stylesheets/components/_phase_banner.scss
```diff
@@ -0,0 +1,33 @@
+$laa-uat-colour: govuk-colour("orange");
+$laa-staging-colour: govuk-colour("turquoise");
+$laa-dev-colour: govuk-colour("pink");
```
naseberry, line 3: Do you need these variables? Couldn't you apply govuk colours directly?

naseberry (approved):
Aside from my question, it looks good to me.

app/webpack/stylesheets/components/_phase_banner.scss
kmahern: @naseberry It would be possible to apply the govuk colours directly but I thought it made sense to use the variables as they are referenced in a couple of places (it is also how it was implemented on FALA https://github.com/ministryofjustice/fala/blob/f126531e74a9787d161b3c28813063a431b27c2b/fala/assets-src/sass/main.scss)

naseberry: @kmahern multiple reference is valid. FALA's implementation works but it is about making ours better. I would suggest the follow:

- Rename  variables from `$laa-` to `$app-`, why? because LAA has no defined colour palette
- If these variables will be used elsewhere, suggest moving them to a separate file, i.e `../stylesheets/settings/_colours.scss`
- MInor issue, use single quotes instead of double quotes.

PS - Hope you don't mind if we copy this feature for CCCD

spec/features/layout_spec.rb
```diff
@@ -1,18 +1,38 @@
...
-      within '.govuk-header__content' do
-        expect(page).to have_link('View court data')
+      within '.govuk-phase-banner' do
+        expect(page).to have_css('.govuk-phase-banner__content', text: 'alpha')
+        expect(page).to have_css('.govuk-phase-banner__text', text: 'This is a new service')
       end
     end
+  end
+
+  context 'when environment is uat' do
```
mabely, line 27: Should we cover the other environments as well in this test?

app/views/layouts/_phase_banner.html.haml
```diff
@@ -1,7 +1,10 @@
-.govuk-phase-banner
+.govuk-phase-banner{ class: ENV['ENV'] }
```
MalcolmVonMoJ, line 1: This new class isn't needed, we can either put it once on the body or the header.  
In FALA, it was on the body, so the CSS would be the same.

app/views/layouts/application.html.haml
```diff
@@ -28,7 +28,7 @@
 
     = render partial: 'layouts/cookie_banner'
 
-    %header.govuk-header{"data-module" => "govuk-header", role: "banner"}
+    %header.govuk-header{"data-module" => "govuk-header", role: "banner", class: ENV['ENV'] }
```
MalcolmVonMoJ, line 31: I am wondering whether we could expand this out a bit, so the class reads "app-environment-dev" (for example).  I assume this would be `class: "app-environment-" + ENV['ENV']` or something like that.  

This makes it more descriptive and less likely to clash with something else should a new class of "dev" or "uat" be added at some point in the future.  

I would have put this on the body tag, but here is okay too.

spec/features/layout_spec.rb
kmahern: 👍 have added tests for the other environments

app/webpack/stylesheets/components/_phase_banner.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.uat {
+  .govuk-header__container {
+    border-bottom-color: $laa-uat-colour;
+  }
+
+  .govuk-tag.govuk-phase-banner__content__tag {
```
MalcolmVonMoJ, line 10: If you keep the class on the header, this should be changed to:

`+ * .govuk-tag.govuk-phase-banner__content__tag {`

This way we don't need the class to be repeated on the phase tag.  

If moving the class to the body, this change is not needed.  

(Same as for both instances of the similar code below)

app/webpack/stylesheets/components/_phase_banner.scss
kmahern: 👍  have renamed variables $app- as suggested and amended to use single quotes. There are no plans to use the variables other than for border-bottom-color and background-color on the govuk-phase-banner so I have left them in the _phase_banner.scss file for now.

app/views/layouts/_phase_banner.html.haml
```diff
@@ -1,7 +1,10 @@
...
   %p.govuk-phase-banner__content
     %strong.govuk-tag.govuk-phase-banner__content__tag
-      = t('generic.alpha')
+      - if %w(uat staging dev).include? ENV['ENV']
+        = ENV['ENV']
+      - else
+        = t('generic.alpha')
```
MalcolmVonMoJ, line 7: Should we set a new variable to the value of `ENV['ENV']` at the top and use that throughout the page.  If none exists (local), we should set it to "local" (or similar).  That way we can also set colours for those running locally.

app/webpack/stylesheets/components/_phase_banner.scss
```diff
@@ -0,0 +1,33 @@
+$laa-uat-colour: govuk-colour("orange");
```
MalcolmVonMoJ: Can we add one in for local.  Perhaps this:
```
$app-local-colour: govuk-colour("bright-purple");
```

app/webpack/stylesheets/components/_phase_banner.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.dev {
+  .govuk-header__container {
+    border-bottom-color: $laa-dev-colour;
+  }
+
+  .govuk-tag.govuk-phase-banner__content__tag {
+    background-color: $laa-dev-colour;
+  }
+}
```
MalcolmVonMoJ, line 33: Can we add one in for local?

MalcolmVonMoJ (changes requested):
Looks good.  I think we need to account for people running locally too, and also, we only need the environment class to appear once in the document, be that in the body or the header doesn't really matter.

app/views/layouts/application.html.haml
MalcolmVonMoJ: See my comment about `ENV['ENV']` below - I suggest setting a new variable to the value and using it throughout.

spec/support/application.rb
```diff
@@ -12,6 +12,14 @@ def fixture_file(file, options = {})
         Rails.root.join(*path)
       )
     end
+
+    def with_env(env)
+      @original_env = ENV['ENV']
```
mabely, line 17: Would it be preferable to use `Rails.env` instead of accessing the environment variables in this way? It would also violate [the Rails/EnvironmentVariableAccess cop](https://docs.rubocop.org/rubocop-rails/2.10/cops_rails.html#railsenvironmentvariableaccess), although we currently override the cop so that we are allowing reads but not writes ([here](https://github.com/ministryofjustice/laa-court-data-ui/blob/fea27b2407dda7a6abface997359b70b248e97ad/.rubocop.yml#L102)).

app/views/layouts/_phase_banner.html.haml
kmahern: 👍 thanks - have moved class to body as suggested

app/views/layouts/application.html.haml
kmahern: 👍 have expanded class name to "app-environment-" as suggested.

app/webpack/stylesheets/components/_phase_banner.scss
kmahern: 👍 have moved class to body

app/views/layouts/_phase_banner.html.haml
kmahern: I have created app_environment in application_helper.

app/webpack/stylesheets/components/_phase_banner.scss
kmahern: 👍 have made suggested change.

app/webpack/stylesheets/components/_phase_banner.scss
kmahern: 👍 have made suggested change.

MalcolmVonMoJ (approved):
LGTM 👍

PR #616, opened 2021-06-16
AAC-103 Added Ruby helper for GDS description lists

MalcolmVonMoJ:
#### What
Simplified the code and added a helper for the GDS description lists

#### Ticket

[AAC-103 Refactor to use helpers to reduce duplication](https://dsdmoj.atlassian.net/browse/AAC-103)

#### Why
Description lists are common components used several times in the bespoke accordion for court applications.  Using a helper makes maintenance easier and the court applications file considerably shorter

#### How
New Ruby helpers for:
- each row in a `<dl>` list
- each key in a row
- each value in a row

The first helper calls the remaining two - which can be called separately if needed on a different page.

### DIFF
--- a/app/helpers/govuk_design_system_helper.rb
+++ b/app/helpers/govuk_design_system_helper.rb
@@ -34,6 +34,29 @@ module GovukDesignSystemHelper
     end
   end
 
+  def govuk_summary_list_entry(key, value, tag_options_key = {}, tag_options_value = {})
+    tag_options_key = prepend_classes('govuk-summary-list__key', tag_options_key)
+    tag_options_value = prepend_classes('govuk-summary-list__value', tag_options_value)
+
+    tag.div do
+      govuk_summary_key(key, tag_options_key).concat(
+        govuk_summary_value(value, tag_options_value)
+      )
+    end
+  end
+
+  def govuk_summary_key(key, tag_options_key)
+    tag.dt(**tag_options_key) do
+      key
+    end
+  end
+
+  def govuk_summary_value(value, tag_options_value)
+    tag.dd(**tag_options_value) do
+      value
+    end
+  end
+  
   private
 
   def prepend_classes(classes_to_prepend, options = {})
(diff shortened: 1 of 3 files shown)


app/helpers/govuk_design_system_helper.rb
```diff
@@ -58,4 +58,27 @@ def page_title(title, caption)
     caption_var = caption.strip.concat if caption.present?
     "#{caption_var} #{page_title_var} - #{service_name} - GOV.UK".strip
   end
+
+  def govuk_summary_list_entry(key, value, tag_options_key = {}, tag_options_value = {})
```
kmahern, line 62: I believe the new govuk_summary_list_entry, govuk_summary_key and  govuk_summary_value methods should all be public (to be consistent with govuk_detail,  govuk_breadcrumb_builder and govuk_page_title).

kmahern: Also, these will all require new tests in govuk_design_system_helper_spec.rb

kmahern (changes requested):
We will need to add tests for the new helper methods.

app/helpers/govuk_design_system_helper.rb
MalcolmVonMoJ: I have moved the methods to the public section and I've added spec tests.

kmahern: Apologies @MalcolmVonMoJ, my mistake, I have had another look at your changes and I think only the govuk_summary_list_entry should be public as this is the only method referenced from outside the class (in _court_applications.html.haml). Can govuk_summary_key and govuk_summary_value please be made private? (if they are public we should also have separate tests for them) Otherwise new test looks good.

kmahern (approved):

app/helpers/govuk_design_system_helper.rb
MalcolmVonMoJ: Done

PR #643, opened 2021-07-07
AAC-168 Notification Banner Helper and Spec Tests

MalcolmVonMoJ:
#### What
Introduces a notification banner helper and spec tests

#### Ticket

[AAC-168 Add Ruby helper for notification banners including a failure option](https://dsdmoj.atlassian.net/browse/AAC-168)

### DIFF
--- a/app/helpers/govuk_design_system_helper.rb
+++ b/app/helpers/govuk_design_system_helper.rb
@@ -47,21 +47,47 @@ module GovukDesignSystemHelper
   end
 
   def govuk_notification_banner(text, key = 'Important', tag_options = {})
-    type_class = ''
-    type_class = 'govuk-notification-banner--success' if key == 'Success'
-    type_class = 'govuk-notification-banner--failure' if key == 'Failure'
-    tag_options = prepend_classes("govuk-notification-banner #{type_class}", tag_options)
-    tag_options[:data] = { module: 'govuk-notification-banner' }
-    banner_title = tag.h2(key, id: 'govuk-notification-banner-title',
-                               class: 'govuk-notification-banner__title')
-    banner_text = tag.p(text, class: 'govuk-body')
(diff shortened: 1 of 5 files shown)


app/helpers/govuk_design_system_helper.rb
```diff
@@ -46,8 +46,35 @@ def govuk_summary_list_entry(key, value, tag_options_key = {}, tag_options_value
     end
   end
 
+  def govuk_notification_banner(text, key = 'Important', tag_options = {})
```
naseberry, line 49: Pattern: I would suggest setting `nil` as the default for the `text` arg

app/webpack/stylesheets/components/_notification_banner.scss
```diff
@@ -0,0 +1,21 @@
...
+    color: govuk-shade($govuk-error-colour, 20%);
+  }
+
+  .govuk-notification-banner__link:active {
+    color: $govuk-error-colour;
+  }
+
+  .govuk-notification-banner__link:focus {
+    color: $govuk-text-colour;
+  }
```
naseberry, line 20: I would recommend using the [govuk-link-style-error mixin](https://github.com/alphagov/govuk-frontend/blob/main/package/govuk/helpers/_links.scss#L130) to keep the style consistent.

app/helpers/govuk_design_system_helper.rb
naseberry: I would consider allowing this method to accept a content block as we may wish to add more than just a paragraph.

See: https://design-system.service.gov.uk/components/notification-banner/success/index.html

MalcolmVonMoJ: Done and done

pking-moj (approved):

PR #646, opened 2021-07-08
Rails 6.1.4 ka

kmahern:
#### What
Update Rails from 6.1.3 to 6.1.4

#### Ticket
https://dsdmoj.atlassian.net/browse/AAC-185

#### Why
This is a dependency update for our applciation framework

#### How

- Updated Rails version in Gemfile
- Ran rails app:update as detailed [here](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) We would not normally need to do this for a minor version update, however, this was not done when we updated to 6.1
- Set the default configuration to 6.1 by setting load_defaults 6.1 as documented [here](https://dsdmoj.atlassian.net/wiki/spaces/CCCD/pages/3496542404/Post+Rails+6.1+upgrade+configuration)
- Added config.action_mailer.deliver_later_queue_name = :mailers as from Rails 6.1 the default fo config.action_mailer.deliver_later_queue_name changes from mailers to nil  and this was causing emails to be added to the default instead of mailers queue (and specs to fail) see [here](https://guides.rubyonrails.org/configuring.html)
- Removed superfluous local: true from form_with calls as the default for Rails 6.1 configuration is now for this to be set to false (see [here](https://guides.rubyonrails.org/configuring.html))

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
```
jsugarman, line 1: we are taking a different approach and stepwise adopting these defaults or not. using separate commits. if you are loading defaults anyway then you should not need this and can get rid of i would have thought

jsugarman (commented):
Too many config changes in this one PR that I am unclear of the effect and aren't explained in the PR itself. We are taking a different approach and being more methodical with the changes whereas you appear to be simply accepting all the new defaults. None of them seem particularly problematic especially since ones like `active_storage` config does not relate to VCD.

One thing, If you are loading all the defaults then you should not need the `new_framework_defaults_6.1.rb` file. 

I guess if you have sanity tested all the major journeys on a hosted environment, and given you have few users (I assume this is still the case), and are willing to accept the risks then I can approve.

config/initializers/new_framework_defaults_6_1.rb
kmahern: @jsugarman thanks, I have removed this file.

jsugarman (approved):
👌

config/environments/development.rb
```diff
@@ -30,54 +32,36 @@
...
-  config.logstasher.log_level = Logger::INFO
-  config.logstasher.controller_enabled = true
-  config.logstasher.mailer_enabled = false
-  config.logstasher.record_enabled = false
-  config.logstasher.view_enabled = false
-  config.logstasher.job_enabled = true
-  config.logstasher.suppress_app_log = false
-  config.logstasher.source = 'laa-court-data-ui-development'
-  config.logstasher.backtrace = true
-  config.logstasher.logger_path = 'log/logstasher_development.log'
```
pking-moj, line 71: Should these lines have been removed? Looks like they are config for logstasher and not rails related.

config/environments/production.rb
```diff
@@ -65,33 +64,25 @@
...
-  config.logstasher.logger = config.logger
-  config.logstasher.controller_enabled = true
-  config.logstasher.mailer_enabled = false
-  config.logstasher.record_enabled = false
-  config.logstasher.view_enabled = false
-  config.logstasher.job_enabled = true
-  config.logstasher.suppress_app_log = true
-  config.logstasher.source = "laa-court-data-ui-production-#{ENV.fetch('ENV', nil)}"
-  config.logstasher.backtrace = false
-  config.logstasher.logger_path = 'log/logstasher.log'
```
pking-moj, line 93: Should these lines have been removed? Looks like they are config for logstasher and not rails related.

config/environments/development.rb
mabely: I've reinstated this as it seems it was removed accidentally

config/environments/production.rb
mabely: I've reinstated this as it seems it was removed accidentally

PR #654, opened 2021-07-16
 Move court-data-adaptor from lib to app/services folder

Obsiye:
### Move court-data-adaptor from lib to app/service folder

This is a concept. I was wondering if it's worth doing this. It leads to 153 deletions and 41 additions. Also, removes that folder with a bunch of requires in it. It makes it easier to understand the code as a new guy (for me at least).

#### Ticket

~~[board ticket description](link)~~ < should i make one of this?

#### Why

- This is because the rails autoloader, zeitwerk, has autoloading capabilities from one of its autoload paths e.g. app/service/
https://guides.rubyonrails.org/autoloading_and_reloading_constants.html
- This means no need for a bunch of require/require_relative calls.

- Another benefit is not needing to require a new court-data-adaptor file every time

- Also, not needing to restart the server whenever a code change is made to the court-adaptor work

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


jsugarman:
@Obsiye this was put in the lib folder as it is more of a re-usable library than a service. 

As a lib it was intended that it could be shared with other apps by moving out to a gem. If you are just wanting to avoid the need to require (and restart server when changes are made) then you can look at [autoloading options](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html)

PR #779, opened 2021-12-02
removed representation orders section from defendants page

ivanELEC:
#### What
Hide the empty representation orders table on VCD. 

#### Ticket

[Hide Representation Orders section in VCD so that caseworkers do not see incomplete information](https://dsdmoj.atlassian.net/browse/AAC-360)

#### Why
This place holder (below screenshot from VCD alpha) was placed into live while we worked out how to expose Representation orders in VCD.

Until we are able to expose the data on VCD we should not be displaying this table on VCD. 

As we are no closer to a decision on bringing data from other systems into VCD, or currently looking to surface the data HMCTS to shares with us (which is information available in our own systems), having this displayed in VCD for our users seems unnecessary.

This ticket is to temporarily hide this from users until we are ready to expose the information in full.

#### How

Approach is to remove the render line in views/defendants/_form.html.haml
This means that once we add this back in again, we still have the representation orders fragment to use (and just need to inject the correct data and add:

`= render 'defendants/representation_orders'`

to views/defendants/_form.html.haml under the line where the defendant fragment is rendered


ivanELEC:
Have tested local, dev, uat, staging
When looking at the defendant page - can see that the Representation Orders aren't there for linked Defendants (and non-linked defendants) with no issues in console log 
No issues on Lighthouse : [20201203_Lighthouse_VCDRemoveRepOrders.pdf](https://github.com/ministryofjustice/laa-court-data-ui/files/7650177/20201203_Lighthouse_VCDRemoveRepOrders.pdf)

Obsiye (approved):

pking-moj (approved):

PR #808, opened 2021-12-29
AAC-263 Upgrade to Ruby 3

Obsiye:
#### What

Upgrade to Ruby 3 (3.0.3) from 2.7.2

#### Ticket

[AAC-263](https://dsdmoj.atlassian.net/browse/AAC-263?atlOrigin=eyJpIjoiNDJmNDJmNzcwZTg4NGI5OThlYjM4NGM4NjNkMGFlMDAiLCJwIjoiaiJ9)

#### Why

[Answer to why can be found here](https://dsdmoj.atlassian.net/browse/AAC-263?atlOrigin=eyJpIjoiNDJmNDJmNzcwZTg4NGI5OThlYjM4NGM4NjNkMGFlMDAiLCJwIjoiaiJ9)

#### How

Update the dockerfile, circleci config, .ruby-version and Gemfile to use ruby 3.0.3

Run bundle which also upgrades some gems and specifies that the Gemfile.lock was upgraded with ruby 3.0.3
(Devs will notice an error when trying to run `rails server`, which should lead them to upgrading their local bundler version)

### **To note:**

You may need to update your bundler to 2.2.3 with `gem install bundler:2.2.3`


ivanELEC (dismissed):

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.7.2-alpine
+FROM ruby:3.0.3-alpine3.15
```
pking-moj, line 1: Is there a reason we are fixing the alpine version? Doing so restricts us from getting patches

Obsiye: Thanks, will change. I didn't realise that was the reason.

pking-moj (approved):

pking-moj (approved):

mabely (approved):
👍🏼

Obsiye:
Approved the additional changes made by @pking-moj

PR #824, opened 2022-01-07
AAC-379 - Fix mailer errors in local development environment 

mabely:
**What**
Whenever a mailer is called on localhost, an error is thrown which is intended but has caused confusion for new developers pertaining to whether it is intended or there is indeed an issue. 

This change allows the mailer to be run without throwing any errors or sending the email, so the user sees the email as having been successfully sent.

**Ticket**
This PR covers 2 tickets:
https://dsdmoj.atlassian.net/browse/AAC-379
https://dsdmoj.atlassian.net/browse/AAC-371

**Why**
1. Missing API key
Emails are sent via govuk_notify_rails gem, which requires an API key, this is not setup for the local dev environment and so it fails for both tickets. 
When a mailer is called, the gem [notifications-ruby-client](https://github.com/alphagov/notifications-ruby-client) (underlying gov Gem that handles mailers) calls `.length` on the API key when it tries to deliver the email, but we are missing the key in our local dev environment.
Applicable to both tickets above.

2. Missing default host for mailers
The first error encountered by [AAC-379](https://dsdmoj.atlassian.net/browse/AAC-379) is when running VCD in localhost and attempting to create any type of user. User creation triggers `devise/mailer`, which calls `set_personalisation`.
`Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true`
The default host is not set for mailers, so when the mailer attempts to call a url helper, it is unable to generate a link without (see commit message for further info).

**How**
1. Developers to add `GOVUK_NOTIFY_API_KEY` to their `.env.development.local`, to be provided separate to this PR
2. Added a default host for mailers


config/environments/development.rb
```diff
@@ -37,6 +37,9 @@
 
   config.action_mailer.perform_caching = false
 
+  # Set default mailer host to be the same as Rails default host to generate urls for mailers.
+  config.action_mailer.default_url_options = { host: 'example.com' }
```
pking-moj, line 41: In `config/environments/test.rb` the value for this is `{ host: 'localhost', port: 3000 }`. Should we perhaps make this the same for consistency sake?

mabely: I agree it would be better to have consistency with the test env, I thought it would only configure the host for action_mailer and it wouldn't be consistent in the rest of the app, but I didn't find any evidence of that

pking-moj (approved):

Obsiye (approved):

PR #833, opened 2022-01-13
AAC-393 - Update ingress to split traffic 50/50

pking-moj:
#### What

As apart of our EKS migration, we need to do a split of traffic to ensure that everything is working as we expect. 

#### Ticket

[Migrate View Court Data Prod onto the Production EKS Implementation](https://dsdmoj.atlassian.net/browse/AAC-393)

#### Why

Mandatory EKS Migration


wordshaker (approved):

ivanELEC:
This change makes sense (splitting traffic of weighted policy 50/50 between production live and production live-1 to ensure everything works as expected)

ivanELEC (approved):

PR #835, opened 2022-01-13
AAC-393 - Make weighting of traffic 100% to live cluster

pking-moj:
What
As apart of our EKS migration, we need to do move the traffic over to the live cluster. 

Ticket
[Migrate View Court Data Prod onto the Production EKS Implementation](https://dsdmoj.atlassian.net/browse/AAC-393)

Why
Mandatory EKS Migration


ivanELEC (approved):
Approving on the basis that nothing went wrong with 50/50 weighting

PR #836, opened 2022-01-13
AAC-162 Cypress Test Cookies and Login/Logout

Obsiye:
#### What

Add Cypress tests for Cookies (banner and settings page) and more Login/Logout tests

#### Ticket

[AAC-162](https://dsdmoj.atlassian.net/browse/AAC-162?atlOrigin=eyJpIjoiNWRjOTdhOTFlYzJkNDQ1ZTllOTc4MmIzZGExNzliYTIiLCJwIjoiaiJ9)

#### Why

- To test and ensure rigidity of the cookies feature
- To not pollute genuine data in Google Analytics
- Add customised data attributes for cypress testing to meet best practices [here](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements)

#### How

- Add cypress tests to accept and reject cookies
- Add cypress tests to adjust cookie settings
- Add cypress testing to check for the change in two cookies: cookies_preferences_set and analytics_cookies_set
- Add cypress tests to test the logout procedure
- Update cypress configuration on all environments to block calls to Google Analytics.
- Add data attributes to buttons/links (accept, reject, settings and form submit)

**TO NOTE:**

Cypress tests run faster for the cookie banner and cookie settings page when the user is not logged in hence i just tested with 'not logged in'. I have tried looking into storing a session for login but its an experimental cypress feature.

This tests cookies whilst not logged in because it's a feature that doesn't currently require authentication

**EDIT:**

These tests only work with JS enabled and doesn't test for the non-js version. This is because cypress requires javascript to work and the current solutions are [workaround / hacks ](https://github.com/cypress-io/cypress/issues/1611#issuecomment-749406961) . If we were to do it, I would suggest a separate ticket / feature.

### DIFF
--- a/cypress/integration/index/cookie_banner.spec.js
+++ b/cypress/integration/index/cookie_banner.spec.js
@@ -4,34 +4,26 @@ describe('Cookie banner', () => {
     cy.checkBanner()
   })
 
-  context('logged in with javascript enabled', () => {
+  context('not logged in and javascript enabled', () => {
     beforeEach(() => {
       cy.visit('/')
-      cy.fixture('users').then((users) => {
-        cy.login(users[0].username, users[0].password)
-      })
     })
 
     const changeCookingSettings = 'Change your cookie settings'
-
-    it('displays search filters page', () => {
-      cy.get('.govuk-fieldset__legend').should(
-        'contain',
-        'Search for'
-      )
-    })
+    const rejectedAdditionalCookies = "You've rejected additional cookies"
+    const hideMessageText = 'Hide this message'
 
     it('can reject cookie preferences', () => {
       cy.getCookie('cookies_preferences_set').should('not.exist')
       cy.checkCookieValue('analytics_cookies_set', 'false')
       cy.get("[data-cy='reject_cookies']")
         .should('contain', 'Reject analytics cookies')
-        .should('have.attr', 'href').and('include', 'analytics_cookies_set=false&show_confirm_banner=true')
+        .and('have.attr', 'href').and('include', 'analytics_cookies_set=false&show_confirm_banner=true')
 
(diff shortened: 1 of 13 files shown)


mabely:
It'd be really helpful if we can test for whether the cookies have been set correctly too

ivanELEC (approved):

mabely:
This is on me for not including it explicitly in the ticket but the behaviour/text in the cookie banner and preferences page is different if javascript is disabled, could this be added into the test or alternatively I can create a separate ticket for it

cypress/integration/index/cookie_banner.spec.js
```diff
@@ -0,0 +1,121 @@
+describe('Cookie banner', () => {
+  before(() => {
+    cy.visit('/')
+    cy.checkBanner()
```
mabely, line 4: Do we want to check this at every test or is it just in the cookies test because there's another banner here at play? 

Also, I feel it'd be helpful to disambiguate this command with `checkEnvBanner` or similar? I thought it was checking the cookie banner in this instance

Obsiye: Makes sense thanks. I'll change the name to checkEnvBanner(), which makes more sense.

As for disabling JS, it was difficult to do but not impossible. The workaround / hack e.g. https://github.com/cypress-io/cypress/issues/1611#issuecomment-749406961 disabled a chrome browser configuration or injection of js. This lead to inconsistencies with how the app actually works, so I just tested with JS enabled. As cypress needed JS to work.

pking-moj (approved):

mabely (approved):

ivanELEC (approved):

PR #877, opened 2022-02-11
AAC-382 - Update heading size and section spacing

pking-moj:
#### What

Based on the prototype app, increasing the heading size and spacing to match this. 

#### Ticket

[Increase sub-heading sizing and spacing between sections in VCD](https://dsdmoj.atlassian.net/browse/AAC-382)

#### Why

Readability of data

#### How

Headings to increase from `govuk-heading-m` to `govuk-heading-l` and add the `govuk-!-margin-bottom-9`


ivanELEC (commented):
Are we considering the changes to the "heading" component to use "govuk-!-margin-bottom-4" a standard throughout the app? Or is the fact that rep-orders doesn't have this additional class indication that these changes are specific to certain pages? 

If the changes to the table and the heading are "standard" could create a helper function instead of manually applying the class each time like we have done for title, detail etc. where we prepend the "govuk-!-margin-bottom-9" class to tables and govuk-!-margin-bottom-4" class to headings

pking-moj:
> Are we considering the changes to the "heading" component to use "govuk-!-margin-bottom-4" a standard throughout the app? Or is the fact that rep-orders doesn't have this additional class indication that these changes are specific to certain pages?
> 
> If the changes to the table and the heading are "standard" could create a helper function instead of manually applying the class each time like we have done for title, detail etc. where we prepend the "govuk-!-margin-bottom-9" class to tables and govuk-!-margin-bottom-4" class to headings

These changes were indicated on the ticket by Dan and thus implemented as required. If this is looked to be standardised i would take it up with him and we can do this in further tickets however the requirements were to match what we have now and fix based on snagging.

PR #878, opened 2022-02-11
AAC-413 Set puma to single mode in dev environment

Obsiye:
#### What

Move Puma to single mode in development environment.

#### Ticket

[AAC-413](https://dsdmoj.atlassian.net/browse/AAC-413?atlOrigin=eyJpIjoiYWRhZWNhN2RhNzhhNDUwZGJiMzZiZTZlOTA4MGRmYTQiLCJwIjoiaiJ9)

#### Why

- This is because in the development environment we are currently running a single worker in Puma's cluster mode, which means we could just opt for the single mode (this means no cluster manager/runner needed)

- This should prove to be minor local development performance improvement.

- Also, removes the log WARNING: Detected running cluster mode with 1 worker.

#### How

- Set Worker count to zero

##### Relevant resources:
https://medium.com/@CGA1123/reducing-rails-memory-usage-by-15-56090b6294bd


ivanELEC (approved):
No issues from me based on understanding that we do not need multiple workers when running VCD

mabely (approved):

pking-moj (approved):

PR #879, opened 2022-02-11
AAC-465 - Add last sign in and export functionality

pking-moj:
#### What

Add last signed in data to the users table and ability to export the table information

#### Ticket

[AAC-465](https://dsdmoj.atlassian.net/browse/AAC-465)

#### Why

Security Requirement

#### How

Expose data from the database we already keep for last signed in. Output data into CSV and output to new endpoint

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -45,7 +45,7 @@ class UsersController < ApplicationController
 
   def export
     response.headers['Content-Type'] = 'text/csv'
-    response.headers['Content-Disposition'] = "attachment;filename=users.csv"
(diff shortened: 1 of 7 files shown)


ivanELEC:
Confirmed that this works when logged in as a manger
Doesn't export file when logged in as a caseworker or admin or not logged in

Minor security risk not having an rspec test for this, shall we make a ticket for this to be handled separately?

ivanELEC (commented):
Could add in tests for other user roles to ensure they can't download the CSV file by going down that path also?

OliverBriggs (approved):

ivanELEC (approved):

wordshaker (approved):

wordshaker (approved):

PR #897, opened 2022-02-24
Dependabot/bundler/prometheus exporter 2.0.0

ivanELEC:
#### What
Selectively bump prometheus export to 2.0.0 (as opposed to 2.0.1)
#### Ticket

[VCD - Release of dependency and security patches (Sprint Wyvern)](https://dsdmoj.atlassian.net/browse/AAC-506)

#### Why
Breaking change [Version 2.0.1 fails to start with NameError on PeriodicStats](https://github.com/discourse/prometheus_exporter/issues/222)

#### How
Set Gemfile so use version 2.0.0

-------


ivanELEC:
> Im not sure what has happened here - but you have updated a whole lot of packages not related to the gem you are dealing with. We maybe need to take a step back and try again

Got the following error "revealed dependencies not in the API" - when researching it, there was a suggestion to remove Gemfile.lock and recreate via bundle command and this is where all the changes came from. Will start from scratch on the selective update after Show and Tell

PR #907, opened 2022-03-04
AAC-452 Apply Landmark Roles to Phase Banner and Breadcrumbs

ivanELEC:
#### Ticket

[VCD - Phase Banner and Breadcrumb not contained within page landmark](https://dsdmoj.atlassian.net/jira/software/c/projects/AAC/boards/622?modal=detail&selectedIssue=AAC-452)

#### Why
Initially discovered from a [moderate WCAG best practise accessibility violation](https://dequeuniversity.com/rules/axe/4.3/region) when doing Cypress-Axe testing

If there is page content not within landmarks users with screen readers and other assistive technology don’t have the ability to skip through to that content using the landmarks, so it might get missed out by them.

#### How
Applied roles to each element that wasn't wrapped in a landmark, in instances where the same role applied to two elements in different sections of the page - followed [ARIA guidelines](https://www.washington.edu/accessibility/web/landmarks/) and applied the landmark to both sections with an associated aria-label that differentiated them based on use/context

- Phase Banner - Complementary role
- Nav Bar - Navigation role, Aria Label (Top Navigation) 
- Breadcrumbs - Navigation role, Aria Label (Case Navigation) 

Notes:

-   Complementary role is supporting info that is associated with main content but can standalone as it's own piece of info. Typically sidebars but can also be things like call out boxes - this may not be the ideal role for this section but it's the most suitable out of the standard ones
-  Top Navigation was used as Aria Label for Nav Bar because it was already the Aria Label for that section, I would have preferred User Navigation as this indicates what is is, links to things user can do that are user specific


mabely (approved):

pking-moj (approved):

app/views/layouts/_navigation.html.haml
```diff
@@ -1,4 +1,4 @@
-%nav.govuk-header__navigation{"aria-label" => t('layouts.application.header.navigation.top_navigation')}
+%nav.govuk-header__navigation{"aria-label" => t('layouts.application.header.navigation.top_navigation'),role:"navigation"}
```
naseberry, line 1: Nits:
- add spacing after the comma
- add spacing after the colon

Coding standard: you're using both hash rocket `attribute => value` and json `attribute: value` syntax style, choose one

lib/gds_design_system_breadcrumb_builder.rb
```diff
@@ -6,7 +6,7 @@
 #
 class GdsDesignSystemBreadcrumbBuilder < BreadcrumbsOnRails::Breadcrumbs::Builder
   def render
-    @context.tag.div(class: 'govuk-breadcrumbs') do
+    @context.tag.div(class: 'govuk-breadcrumbs', role: 'navigation', 'aria-label': 'Navigate Case') do
```
naseberry, line 9: Nit: Could the arial-label value be internationlized using `I18n.t()`?

naseberry:
Looks good aside from the minor nits. 👍

Obsiye (approved):

app/views/layouts/_navigation.html.haml
ivanELEC: Noted, this inconsistency is throughout the app, affecting 11 HAML files - will update this instance but also might be a sign that we need to review our HAML linting (or implement it) - making a ticket for this this morning

Obsiye (approved):

lib/gds_design_system_breadcrumb_builder.rb
ivanELEC: Will add a tech debt ticket for this as there are many other similar examples in different parts of the code where a string is used instead of locale text

PR #938, opened 2022-04-06
AAC-556 - Add MAAT linking to laa_references_controller

pking-moj:
#### What

Add in the usage of V2 calls to CDAPI when the feature flag is set to true.

#### Ticket

[Implement MAAT Linking against CDAPI](https://dsdmoj.atlassian.net/browse/AAC-556)

#### Why

This is apart of the integration piece for V2 of CDA/CDAPI

#### How


app/controllers/laa_references_controller.rb
```diff
@@ -80,13 +80,41 @@ def link_attempt_params
...
+        Sentry.capture_exception(@laa_reference.errors)
+        render_new(I18n.t('laa_reference.link.failure'), I18n.t('error.it_helpdesk'))
+      else
+        redirect_to_edit_defendants
+      end
+    else
+      logger.info 'CALLING_V1_MAAT_LINK'
+      @laa_reference.save
+      redirect_to_edit_defendants
+    end
```
Obsiye, line 115: Implementation looks really good to me.

The below are suggestions  but i'm not great at OOP design in rails, so no offence taken if you prefer not to do the below or something else.

Maybe an addition could be service objects.

Like this method can call a service object or service objects for more readability, reusability, separation of concerns and ease of testing etc.
https://www.honeybadger.io/blog/refactor-ruby-rails-service-object/ 
e.g. 

```
def resource_save
   if Feature.enabled?(:laa_references)
     CDAPI::V2::LaaReference.call(resource: @laa_reference)
     # will our api just be V2? or is it YAGNI...e.g. version 3
   else
     CourtDataAdaptor::Resource::LaaReference
   end
end

in another file ------

class CDAPI::V2::LaaReference
  def self.call(laa_reference)
    new(laa_reference).call
  end
  
  def call(resource)
    laa_reference.save!
  end
  
end
```

Maybe there's a design pattern to place error handling in the `class base_model < ActiveResource` like abstract factory (i think)


Also for logging, i thought `ActiveResource::Base.logger = ActiveRecord::Base.logger
` would be enough to at least log what we're calling? If we use CDAPI::V2 in the class name...

app/controllers/laa_references_controller.rb
```diff
@@ -80,13 +80,41 @@ def link_attempt_params
...
 
   def resource_params
     @resource_params ||= @link_attempt.to_link_attributes
   end
 
+  def resource_save
+    if Feature.enabled?(:laa_references)
+      begin
+        logger.info 'CALLING_V2_MAAT_LINK'
+        @laa_reference.save!
```
Obsiye, line 100: a shebang will mean activeresource validations always run before the call is sent out. Although, we don't have validations on laa_reference...should we? 

I don't think using save! is a problem because at least it says we always run validations before sending to the api in our app... it's easy to miss the difference between save and save! so using save! everywhere protects us from missing validations...
i don't mind if it's save! everywhere instead of patches of save and save! throughout our app
 idk, this is a stream of thought i'm having now sorry

Obsiye, line 100: a shebang will mean activeresource validations always run before the call is sent out. Although, we don't have validations on laa_reference...should we? 

I don't think using save! is a problem because at least it says we always run validations before sending to the api in our app... it's easy to miss the difference between save and save! so using save! everywhere protects us from missing validations...
i don't mind if it's save! everywhere instead of patches of save and save! throughout our app
 idk, this is a stream of thought i'm having now sorry

app/models/court_data_api_resource.rb
```diff
@@ -0,0 +1,23 @@
...
+  def site
+    set_my_active_record_fields unless active_record_fields_set
+    super
+  end
+
+  def set_my_active_record_fields
+    self.site = Rails.configuration.x.court_data_api_config.uri
+  end
+end
+
```
ivanELEC, line 22: Not sure what the purpose of this code is, would you mind describing it?

pking-moj: When we are testing against ActiveResource it has trouble pulling environment variables at the right time to set the site field and will through an exception. This Module allows for the site to be set in a timely fashion so that it will always work

app/models/court_data_api_resource.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# rubocop:disable Style/OptionalBooleanParameter
+
+module CourtDataApiResource
```
mabely, line 5: Might be helpful to add a comment here to state the purpose of the module

mabely (approved):

ivanELEC (approved):

PR #967, opened 2022-04-28
AAC-606 - Add hearing events from V2 if flag enabled

pking-moj:
#### What

Add Hearing Events to come from V2 if hearing_events env var is enabled. 

#### Ticket

[Integrate CDAPI Hearing Events to the Hearing Page events table](https://dsdmoj.atlassian.net/browse/AAC-606)

#### Why

Migration to V2 for View Court Data

#### How

- Create new v2 hearing_events partial as we no longer need a decorator to sort through the prosecution_cases response
- Add model to hit hearing_events endpoint
- Add logic into hearings controller to call hearing_events if V2 enabled
- Add logic in hearing page to show correct table based on V2 flag

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -5,7 +5,8 @@ require_dependency 'court_data_adaptor'
 class HearingsController < ApplicationController
   before_action :load_and_authorize_search,
                 :set_hearing,
-                :set_hearing_day
+                :set_hearing_day,
+                :set_hearing_events
 
   add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
   add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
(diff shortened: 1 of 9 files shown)


Obsiye (approved):
LGTM. To note, there are no feature tests for V2 here but i don't think we have feature tested hearing events before either. It'd be a nice to have in this PR or a future tech debt

pking-moj:
> LGTM. To note, there are no feature tests for V2 here but i don't think we have feature tested hearing events before either. It'd be a nice to have in this PR or a future tech debt ticket

I agree - i have added tests based on existing testing but i think once we are in a good place with a single source of truth for our calls having better testing is key

ivanELEC (approved):

ivanELEC:
> > LGTM. To note, there are no feature tests for V2 here but i don't think we have feature tested hearing events before either. It'd be a nice to have in this PR or a future tech debt ticket
> 
> I agree - i have added tests based on existing testing but i think once we are in a good place with a single source of truth for our calls having better testing is key

agree - if, for this reason, our code climate coverage was below thresholds - we would attack it now, but might be better for us to re-evaluate in general how we're doing some of these tests. There's a huge opportunity to make tests clearer when we switch to v2 fully

PR #984, opened 2022-05-09
Aac 560 integrate cdapi hearingsummaries paginator

ivanELEC:
#### Ticket

[Integrate CDAPI HearingSummaries to Hearings Page (Paginator)](https://dsdmoj.atlassian.net/browse/AAC-560)

#### Why
We are moving to use v2 of Court Data Adaptor (CDA), for this reason, we have created an API (CDAPI) to decouple the data processing from the front-end when linking to the new version of CDA.

This story is for the pagination of the Hearings page, to ensure that a use can see the relevant date title of the page and move between hearing days via the pagination buttons


app/helpers/application_helper.rb
```diff
@@ -36,14 +38,13 @@ def hearings_sorter_link(prosecution_case, column, title = nil)
   end
 
   def app_environment
-    "app-environment-#{ENV.fetch('ENV', 'local')}"
+    "app-environment-#{ENV.fetch('ENV', nil) || 'local'}"
```
pking-moj, line 41: What difference does this change make?

app/controllers/hearings_controller.rb
```diff
@@ -16,7 +16,7 @@ class HearingsController < ApplicationController
   def show
     add_breadcrumb "#{t('generic.hearing_day')} #{hearing_day&.strftime('%d/%m/%Y')}", ''
 
-    return if hearing
+    return if @hearing
```
pking-moj, line 19: Does this need to be an instance variable?

app/helpers/application_helper.rb
ivanELEC: This was a Rubocop automatic correction: 
image
This is the reason, quoted from [here](https://docs.rubocop.org/rubocop/cops_style.html#references-37): 

> This cop suggests ENV.fetch for the replacement of ENV[]. ENV[] silently fails and returns nil when the environment variable is unset, which may cause unexpected behaviors when the developer forgets to set it. On the other hand, ENV.fetch raises KeyError or returns the explicitly specified default value.

app/controllers/hearings_controller.rb
ivanELEC: I used the instance variable because there is also a hearing method, this hearing method is what's used for the v1 call, whereas the @hearing instance variable is the result of either the v1 or v2 call depending on the flag

app/helpers/application_helper.rb
ivanELEC: oh you're right - I must have been running against old code as you said, reverting back to:
`app-environment-#{ENV.fetch('ENV', 'local')}`

config/database.yml
```diff
@@ -82,4 +82,4 @@ production:
   <<: *default
   database: laa_court_data_ui_production
   username: laa_court_data_ui
-  password: <%= ENV['LAA_COURT_DATA_UI_DATABASE_PASSWORD'] %>
```
pking-moj, line 85: Can we not change this file?

app/helpers/application_helper.rb
```diff
@@ -49,7 +49,6 @@ def v2_hearing_summaries?
 
   def decorator_instance(object, decorator_class = nil)
     return object if object.is_a?(BaseDecorator)
-
```
pking-moj, line 52: Does this file need to be changed?

spec/support/webmock/court_data_api/webmock.rb
```diff
@@ -248,4 +248,44 @@
...
+  end
+
+  config.before(:each, stub_hearing_summaries_v2: true) do
+    stub_request(:get, %r{http.*/v2/hearingsummaries/})
+      .to_return(
+        status: 200,
+        body: load_json_stub('cd_api/hearingsummaries_body.json'),
+        headers: { 'Content-Type' => 'application/json' }
+      )
+  end
```
pking-moj, line 278: Should this stub be here? The request url is incorrect

.rubocop.yml
```diff
@@ -191,4 +191,4 @@ Rails/I18nLocaleTexts:
     - spec/**/*.rb
 
 RSpec/VerifiedDoubleReference:
-  Enabled: false
```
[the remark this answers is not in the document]
ivanELEC: The cop is for enforcing a consistency on verified double referencing, it's either referenced as a class or referenced as a string that represents what class it should be. For all instances where there's an instance double created for ResultSet in tests, it fails when referencing the class because the class doesn't exist. I'm not actually sure what is being referenced here to be honest - ResultSet seems to be a catch all for a set of results that comes from JSON API Client. Assuming this cop was caught out due to a rubocop update as I haven't changed a lot of the tests that fail when this line is disabled

spec/support/webmock/court_data_api/webmock.rb
ivanELEC: resolved

pking-moj (approved):

Obsiye (approved):

PR #985, opened 2022-05-09
AAC-559 Integrate CDAPI HearingSummaries to Prosecution Case Page

Obsiye:
#### Ticket

[AAC-559](https://dsdmoj.atlassian.net/browse/AAC-559)

#### Why

We are moving to use V2 of Court Data Adaptor (CDA) via CDAPI

#### How

> **From the CDAPI hearingsummaries response:**
> The defendants table uses overall_defendants data
> The hearing summaries table uses hearing_summaries data
> The breadcrumb uses the prosecution_case_reference

- Add feature_flag hearing_summaries / ENV["HEARING_SUMMARIES"]
- Add CaseSummary ActiveResource class that calls CDAPI `/hearingsummaries` endpoint
- Add Decorators to allow the ActiveResource models data to be represented on the view just like V1

Add exception handling to the prosecution case controller, so that any active resource exceptions on v2 sends the user to the search page ⚠️ (to be further discussed)

- Add authorisation on calls to CDAPI using the activeresource models and the cancancan gem.
- Add a version2 parameter to the Search model to decouple it from a specific feature flag because the model is used on multiple controllers/pages
- Ignore Rubocop cop `Rails/HasManyOrHasOneDependent:` for cd_api models because the associations are on ActiveResource and not on ActiveRecord models which the cop expects.

- Build the name for the DefenceCounselDecorator in hearing summaries and OverallDefendantDecorator in overall_defendants by concatenating the first_name, middle_name, last_name attributes on v2 (as opposed to v1 which had name attribute passed instead of portions of a name).

- Add Todo for the logic of attendance days for each defence counsel on the hearings table - this is more to do with multi day hearings. As this is not in V1, haven't done this for V2 as it's expected to be tech debt.

 
### Testing

1. Add request test, model tests, decorator tests and view tests
2. Add **FactoryBot** for CDAPI models to make it easier to create model objects in tests. Also provides a central place describing the attributes cd_api models contain for testing. 

--------

#### TODO (wip)

 - [X] Discuss Exception handling for this
 - [ ] Create confluence document of Tech Debt found in this PR with explanations / descriptions. ( e.g. name builder, exception handling)
 - [X] Use prosecution_case_reference from v2 response

---------
#### Exception handling of V2 ActiveResource atm sends the user to the search results page using the same urn
image


app/decorators/cd_api/defence_counsel_decorator.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    private
+
+    def name_and_status_blank?
+      name.blank? && status.blank?
+    end
+
+    def name
+      # TODO: # TODO: Create a name service to build the name for reusability
```
pking-moj, line 18: nitpit: double TODO:

app/decorators/cd_api/hearing_day_decorator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module CdApi
+  class HearingDayDecorator < BaseDecorator
+    def to_datetime
+      sitting_day&.to_datetime
```
pking-moj, line 6: Opinion: Does this need to be a decorator for a single method? the testing/complexity vs actually writing this seems overkill.

Obsiye: Ah yep, i'm hoping to write that as tech debt.

The HearingsSorter and HearingsDateSorter  is where it expects the hearing_day to be a string that can convert to DateTime.

In V2 it's an object CdApi::HearingDay with an attribute sitting_day. I added a decorator to allow the call #to_datetime on the object and return siting_day in DateTime type because this object just has one attribute sitting_day.

I'll write this up as tech debt to fix. The alternative will bloat this PR more than having the decorator and a tech debt ticket

pking-moj (approved):

PR #1026, opened 2022-05-30
Aac 656 accessing defendant page via search

ivanELEC:
#### What
[Accessing Defendant Page via Search Fails in v2](https://dsdmoj.atlassian.net/jira/software/c/projects/AAC/boards/622?modal=detail&selectedIssue=AAC-656)

#### Why
Because when accessing defendant page via search page when search page is in v2, opening the page fails as the linked? method is missing

#### How
Add linked method to CD API defendant model using the first offence summary

### DIFF
--- a/app/models/cd_api/defendant.rb
+++ b/app/models/cd_api/defendant.rb
@@ -3,7 +3,7 @@
 module CdApi
   class Defendant < BaseModel
     def linked?
-      self.offence_summaries[0]&.laa_application&.reference&.present?
+      offence_summaries[0]&.laa_application&.reference&.present?
     end
   end
 end
--- a/app/models/cd_api/defendant.rb
+++ b/app/models/cd_api/defendant.rb
@@ -2,5 +2,8 @@
 
 module CdApi
   class Defendant < BaseModel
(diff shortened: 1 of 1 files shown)


app/models/cd_api/defendant.rb
```diff
@@ -2,5 +2,9 @@
 
 module CdApi
   class Defendant < BaseModel
+    def linked?
+      maat_references = offence_summaries.map { |offence| offence&.laa_application&.reference }
+      maat_references.compact.first.present?
```
jrmhaig, line 7: I don't know how computationally intensive it is to find `offence&.laa_application&.reference` but if there is a significant cost (e.g. if there are lots of `offence_summaries` and `laa_applcation` has to perform a look-up) you can use a `lazy` enumerable. For example,

```suggestion
      maat_references = offence_summaries.lazy.map { |offence| offence&.laa_application&.reference }
      maat_references.any(&:present?)
```

This will stop looking for references when it has found one.

jrmhaig (approved):

wordshaker (approved):

PR #1043, opened 2022-06-09
Update page to deal with empty hearing

pking-moj:
#### What

Enable the pagination of hearings where there isnt a hearing to show. This enables the user to still view a "hearing" that doesnt exist but is being shown in the summaries table.

#### Ticket

[V2 Hearing Day Page Fix](https://dsdmoj.atlassian.net/browse/AAC-691)

#### Why

V2 Migration

#### How

- Return an empty hearing instead of returning nothing and a flash, then redirecting
- Update the pages to adjust for the fact the hearing may not have any properties in there
- Update tests accordingly


app/controllers/hearings_controller.rb
```diff
@@ -61,18 +61,19 @@ def set_hearing_events
...
   def hearing_v2_call
     @hearing ||= helpers.decorate(
-      CdApi::Hearing.find(params[:id], params: {
+      CdApi::Hearing.find(hearing_id, params: {
                             date: paginator.current_item.hearing_date.strftime('%F')
                           }), CdApi::HearingDecorator
     )
   rescue ActiveResource::ResourceNotFound
-    redirect_to_prosecution_case(notice: I18n.t('hearings.show.flash.notice.no_hearing_details'))
+    # Return empty hearing so we can still display the page
```
Obsiye, line 75: Is it worth monitoring this on Sentry with `Sentry.capture_exception(e)` ?

Obsiye (approved):

app/controllers/hearings_controller.rb
pking-moj: This isnt an exception as such so no need to capture in Sentry. The hearing summaries can show hearings that were never hearings (things that were scheduled but never heards, future hearings etc) so there is a valid reason for them being on the summaries but their not being a hearing for them.

ivanELEC (dismissed):

ivanELEC (dismissed):

SamKyprianou (dismissed):

ivanELEC:
Tested locally earlier today with V2 flags enabled, no issues on my end so approving

ivanELEC (approved):

SamKyprianou (approved):

app/views/hearings/show.html.haml
```diff
@@ -9,12 +9,13 @@
     - unless Feature.enabled?(:hearing)
       = render partial: 'details', locals: { hearing: @hearing, hearing_day: @hearing_day }
     - else
-      = render partial: 'details_v2', locals: { hearing: @hearing, hearing_day: @hearing_day }
+      = render partial: 'details_v2', locals: { hearing_summary: @prosecution_case.hearing_summaries.find { |each| each.id == @hearing_id },
```
Obsiye, line 12: nit picky sorry. could `|each|` be renamed to `|hearing_summary|`

Obsiye (approved):

PR #1045, opened 2022-06-10
AAC-180 BugFix Broken redirect link when referrer is /searches page

Obsiye:
#### What

If the http_referrer on the cookies page was from /searches page then rebuild the searches link
 using current_search_params found in applicaiton_controller applicaiton_controller.

This is a temporary fix as redirecting back to /searches fails with a 500 exception rather than /searches/new?... which is the more accurate link
for that page.

#### Ticket

[AAC-180](https://dsdmoj.atlassian.net/browse/AAC-180)

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2


ivanELEC (approved):

SamKyprianou (approved):
Looks good to me

pking-moj (approved):
Nice fix 🥇

PR #1064, opened 2022-06-24
AAC-746 BugFix cookie banner breaks on pages that require query params

Obsiye:
#### What

AAC-746 BugFix cookie banner breaks on pages that require query params.

#### Ticket

[VCD - Application crash accepting cookies on defendant page](https://dsdmoj.atlassian.net/browse/AAC-182)

#### Why

The Cookie banner had links that used `href: '?'` and `href=?analytics_cookies_set=true...` which removed the current pages query params such as the param[:id] for the defendant page causing an automatic refresh to fail with a 500 exception.

The solution was to prevent the need for a refresh and [Turbo ](https://github.com/hotwired/turbo-rails) was the best alternative to a custom made JavaScript solution.

#### How

- Implement Turbo (turboframes and turbostreams) to update the banner rather than having to refresh the page (which reduces the number of calls to CDA).

- Due to the no refresh using Turbo,  render the partial 'layouts/google_analytics' in the Turbo_success partial. This will set the google analytics cookies if the user has accepted its usage.

- Use` "data-turbo" => false` where **Devise** authentication is involved because of it **not working well with Turbo**. Without `"data-turbo" => false` the flash notifications fail to work relating to Devise.

co-author: @mabely

⚠️  **To note:**

This creates a new route with endpoint /update_cookies, which is available without sign-in.

**TODO:** 
 - [ ] Document the decision to choose Turbo over other solutions.


pking-moj (approved):
Tested in UAT and happy this works as intended

ivanELEC (approved):

SamKyprianou (approved):

PR #1066, opened 2022-06-24
Aac 715 make cd api endpoints restful: hearing events

ivanELEC:
#### What
UI now uses /hearings instead of /hearing when trying to get a hearing
UI now /hearings/{hearing_id}/hearing_event instead of /hearing_event when trying to get a hearing's hearing events

#### Ticket

[Make CD API Endpoints RESTFUL](https://dsdmoj.atlassian.net/browse/AAC-715)

#### Why
So that our APIs are more in line with RESTFUL API design


SamKyprianou (approved):

pking-moj (approved):

app/models/cd_api/hearing_events.rb
```diff
@@ -2,5 +2,9 @@
 
 module CdApi
   class HearingEvents < BaseModel
+    def self.find(id, params)
```
Obsiye, line 5: This method seems to be tied to hearing_events

Could we rename this to `find_hearing_events` or make the method with > endpoint from: "/v2/...../hearing_events" dynamically change

Obsiye (approved):
LGTM. I left a comment on refactoring a method but it's not a blocker

app/models/cd_api/hearing_events.rb
ivanELEC: Not sure there would be sufficient reasoning for this. In this instance, we are overwriting the existing find function that HearingEvents inherits from the BaseModel. Changing the method name would lead to inconsistency in our controllers - where normally we just call find, in this instance we would be calling HearingEvents.find_hearing_events

PR #1085, opened 2022-06-30
Aac 703 fix haml linting instance variables

ivanELEC:
#### Ticket

[Fix HAML Linting Issues: InstanceVariables](https://dsdmoj.atlassian.net/browse/AAC-703)

#### Why
To fix linting violation "InstanceVariables" which denotes not to use Instance Variables in partials
This is mainly because partials should be considered as "components" that can be re-used. If you use an instance variable for a partial, you lost the ability to use different points of data for different usages of the partial, thus fixing it's usage instead of allowing it to act as a component. More [here](http://www.carlosramireziii.com/stop-using-instance-variables-in-partials.html#:~:text=It's%20tempting%20to%20take%20the,the%20local%20data%20each%20time.)

#### How
In any usage of instance variable in a partial, pass the variable into the partial via locals instead


ivanELEC:
@dependabot recreate

ivanELEC:
@dependabot rebase

PR #1092, opened 2022-07-04
AAC-708 Fix haml-lint MultilinePipe warning

Obsiye:
#### Ticket

[AAC-708](https://dsdmoj.atlassian.net/browse/AAC-708)

#### Why

Maintenance of our service with good coding practices that are standardised with the haml-lint package.

#### How

- Run `haml-lint -i MultilinePipe`
- Fix error warning in code

### DIFF
--- a/.haml-lint_todo.yml
+++ b/.haml-lint_todo.yml
@@ -44,6 +44,7 @@ linters:
       - "app/views/prosecution_cases/cd_api/_raw_response.html.haml"
 
   # Offense count: 1
-  MultilinePipe:
+  IdNames:
     exclude:
       - "app/views/devise/shared/_error_messages.html.haml"
+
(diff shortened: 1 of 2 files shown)


ivanELEC (approved):

ivanELEC:
As stated in Slack, there is ample justification to remove the _error_messages partial from the app, this is because it is a leftover generated file from devise which shows the user information they don't need to see - all pathways of errors and issues via the user management process are already handled by us with error messages decided by the team, having these devise error messages would be confusing to the user

ivanELEC:
> As stated in Slack, there is ample justification to remove the _error_messages partial from the app, this is because it is a leftover generated file from devise which shows the user information they don't need to see - all pathways of errors and issues via the user management process are already handled by us with error messages decided by the team, having these devise error messages would be confusing to the user

ticket has been made to do this and added to back pocket log for another sprint. Approving

ivanELEC (approved):

pking-moj (approved):

PR #1093, opened 2022-07-04
Aac 707/fix haml linting id names

SamKyprianou:
#### What
fixing the ID names to lisp case in HAML files to remove linting issues

#### Ticket

[AAC-707 Fix HAML Linting Issues: IdNames](https://dsdmoj.atlassian.net/browse/AAC-707)

#### How

Ran haml-lint -i IdNames and fixed all errors displayed - then removed IdNames from linting todo list and tested for any hidden issues.


ivanELEC (commented):
As stated in Slack, there is ample justification to remove the _error_messages partial from the app, this is because it is a leftover generated file from devise which shows the user information they don't need to see - all pathways of errors and issues via the user management process are already handled by us with error messages decided by the team, having these devise error messages would be confusing to the user

ivanELEC:
> As stated in Slack, there is ample justification to remove the _error_messages partial from the app, this is because it is a leftover generated file from devise which shows the user information they don't need to see - all pathways of errors and issues via the user management process are already handled by us with error messages decided by the team, having these devise error messages would be confusing to the user

ticket has been made for the above and moved to back pocket log to deal with later - approving

ivanELEC (approved):

PR #1152, opened 2022-08-16
aac-781 MAAT ref check and view change

SamKyprianou:
#### What

CDUI V2 - Do not display dummy MAAT references prefixed with 'Z'

#### Ticket

[AAC-781](https://dsdmoj.atlassian.net/browse/AAC-781)

#### Why

We should ignore maat references prefixed with 'Z' and assume the Defendant is not linked. The reason these references exist is due to a limitation in Common Platform with being unable to remove MAAT references, the Z reference is a placeholder to show removal - to show this to our users would cause confusion regarding whether a case has legal aid granted and has been correctly linked between MLRA and Common Platform.

#### How

Added a check in the linked? method in orverall_defendant_decorator to check for the first character of the maat_references and returning false if it starts with "Z" and directing user to the correct endpoint (laa_references).

`maat_reference&.present? && maat_reference.first != "Z"`

Added check in overall_defendant view page to return "not linked" if maat_reference begins with "Z"

`= overall_defendant.linked? ?  overall_defendant.maat_reference : t('search.result.defendant.unlinked')`

Added check in defendant model to link the defendant names to the correct end point on the searches page.

```
def linked? 
     maat_references.first.present? && maat_references.first.first != 'Z'
    end

    private

    def maat_references
      offence_summaries.filter_map { |offence| offence&.laa_application&.reference }
    end
```

**Test Spec**

Checking the return values of maat_references beginning with "Z".

`   context 'when maat_reference exists and does not begin with "Z"' do
      let(:overall_defendant) { build :overall_defendant, maat_reference: '1234567' }

      it 'returns true' do
        expect(call).to be_truthy
      end
    end

    context 'when maat_reference exists and begins with "Z"' do
      let(:overall_defendant) { build :overall_defendant, maat_reference: 'Z1234567' }

      it 'returns false' do
        expect(call).to be_falsey
      end
    end`

### DIFF
--- a/app/decorators/cd_api/overall_defendant_decorator.rb
+++ b/app/decorators/cd_api/overall_defendant_decorator.rb
@@ -8,7 +8,7 @@ module CdApi
     end
 
     def linked?
-      (maat_reference&.present? && maat_reference.first != "Z") ? maat_reference&.present? : false
+      maat_reference&.present? && maat_reference.first != "Z"
     end
   end
 end
--- a/app/decorators/cd_api/overall_defendant_decorator.rb
+++ b/app/decorators/cd_api/overall_defendant_decorator.rb
@@ -8,7 +8,7 @@ module CdApi
     end
 
     def linked?
-      maat_reference&.present?
+      (maat_reference&.present? && maat_reference.first != "Z") ? maat_reference&.present? : false
     end
   end
 end
--- a/app/decorators/cd_api/overall_defendant_decorator.rb
+++ b/app/decorators/cd_api/overall_defendant_decorator.rb
@@ -8,7 +8,7 @@ module CdApi
     end
 
     def linked?
-      maat_reference&.present? && maat_reference.first != "Z"
+      maat_reference&.present? && maat_reference.first != 'Z'
     end
   end
 end
--- a/app/models/cd_api/defendant.rb
+++ b/app/models/cd_api/defendant.rb
@@ -2,9 +2,16 @@
 
 module CdApi
   class Defendant < BaseModel
+    has_many :offence_summary, class_name: 'cd_api/offence_summary'
+
     def linked?
-      maat_references = offence_summaries.map { |offence| offence&.laa_application&.reference }
-      (maat_references.compact.first.present? && maat_references.compact.first.first != "Z") ? maat_references.compact.first.present? : false
+      maat_references.first.present? && maat_references.first.first != 'Z'
+    end
+
+    private
+
+    def maat_references
+      offence_summaries.map { |offence| offence&.laa_application&.reference }.compact
     end
   end
 end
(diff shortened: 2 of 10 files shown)


app/views/prosecution_cases/cd_api/_overall_defendants.html.haml
```diff
@@ -16,4 +16,4 @@
         %td.govuk-table__cell
           = l(overall_defendant.date_of_birth&.to_date)
         %td.govuk-table__cell
-          = overall_defendant.maat_reference || t('search.result.defendant.unlinked')
+          = overall_defendant.maat_reference.first != "Z" ?  overall_defendant.maat_reference : t('search.result.defendant.unlinked')
```
pking-moj, line 19: Would the below not work? The logic sits in the decorator for this already.
```suggestion
          = overall_defendant.linked? ?  overall_defendant.maat_reference : t('search.result.defendant.unlinked')
```

app/decorators/cd_api/overall_defendant_decorator.rb
```diff
@@ -8,7 +8,7 @@ def name
     end
 
     def linked?
-      maat_reference&.present?
+      (maat_reference&.present? && maat_reference.first != "Z") ? maat_reference&.present? : false
```
Obsiye, line 11: try this
```suggestion
      maat_reference&.present? && maat_reference.first != "Z"
```

app/models/cd_api/defendant.rb
```diff
@@ -4,7 +4,7 @@ module CdApi
   class Defendant < BaseModel
     def linked?
       maat_references = offence_summaries.map { |offence| offence&.laa_application&.reference }
-      maat_references.compact.first.present?
+      (maat_references.compact.first.present? && maat_references.compact.first.first != "Z") ? maat_references.compact.first.present? : false
```
Obsiye, line 7: Could this be unit tested?

I see that there's no model test for this e.g. app/models/cd_api/defendant_spec.rb

I'm not even sure if this model#linked? method is being used 🤔

SamKyprianou: This model#linked? method is used to identify the endpoint for the linked names in the search search results page so i think its definitely used.

Maybe we need to discuss creating a test spec for the model?

app/views/prosecution_cases/cd_api/_overall_defendants.html.haml
SamKyprianou: Ahhh yeah - this makes sense to use the decorator method. Tested locally and its still works as should. will commit suggestion

ivanELEC (approved):

Obsiye (approved):

pking-moj (approved):

PR #1481, opened 2023-04-29
Fix decorator to use pleas array

pking-moj:
#### What

Fix the decorator to use the pleas array rather than the initial object when we were implementing this. 

#### Ticket

N/A

### DIFF
--- a/app/decorators/offence_summary_decorator.rb
+++ b/app/decorators/offence_summary_decorator.rb
@@ -2,11 +2,21 @@
 
 class OffenceSummaryDecorator < BaseDecorator
   def plea_sentence
-    return t('generic.not_available') if plea&.value.blank?
-    return_plea plea
+    return t('generic.not_available') if pleas.blank?
+    return pleas unless pleas.is_a?(Enumerable)
+
+    safe_join(plea_sentences, tag.br)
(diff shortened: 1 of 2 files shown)


app/decorators/offence_summary_decorator.rb
```diff
@@ -2,12 +2,22 @@
 
 class OffenceSummaryDecorator < BaseDecorator
   def plea_sentence
-    return t('generic.not_available') if plea&.value.blank?
-    return_plea plea
+    return t('generic.not_available') if pleas.blank?
+    return pleas unless pleas.is_a?(Enumerable)
+
+    safe_join(plea_sentences, tag.br)
   end
```
Obsiye, line 10: Looks good to me. I noticed that there's no unit test for the plea sentence / OffenceSummaryDecorator

ivanELEC (approved):

app/decorators/offence_summary_decorator.rb
pking-moj: Its integrated tested within the multiple different tests we have so i dont think we need to go to the nth degree testing this.

Obsiye (approved):

PR #1534, opened 2023-06-02
CTSKF-439 - Set MAAT username max length to 10

pking-moj:
#### What

Adds a max length to the MAAT username as this cannot be longer than 10

#### Ticket

[Improve handling of too long usernames](https://dsdmoj.atlassian.net/browse/CTSKF-439)

#### Why

MAAT usernames by design are a max length of 10. If we input anything above 10 chars at the moment we get an error that doesnt match what is happening. This prevents users from actually going over the limit.

#### How

Set the inputs max length to 10


Obsiye (approved):
we could also add this to the User model as validation and thus an error message would pop up if they went above the length e.g. Username must be less than 11 characters long.

However, having this in the UI may be enough feedback to the user.

I'll approve though because it does solve the problem 

EDIT: I just saw the DoD of the ticket and this change looks good to me

ivanELEC (approved):

PR #1554, opened 2023-06-12
CTSKF-449 - Add Judicial Results to Hearing Page

pking-moj:
#### What

Added Judicial Results behind a flag to the hearings page to improve the experience for our users

#### Ticket

[As a caseworker I need to see sentencing details and court orders so that I can authorise the correct payment](https://dsdmoj.atlassian.net/browse/CTSKF-449)

#### Why

Case at a glance is what we are trying to achieve for MVP. This partial takes us one step closer

#### How

- Create Partial for Judicial Results
- Provide required decorators
- Put behind feature flag
- Update and add tests


app/decorators/cd_api/judicial_results_decorator.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    private
+
+    def map_judicial_result_prompts
+      prompts.flat_map
+    end
+
+    def filter_judicial_results_prompts(prompt)
+      Rails.application.config.x.judicial_results['filter'].any? do |filter|
```
Obsiye, line 23:
```suggestion
      Rails.application.config.x.judicial_results.filter.any? do |filter|
```

PR #1778, opened 2023-11-09
CTSKF-615 BugFix Alert box not appearing in tests after deleting a user

Obsiye:
#### What

Wait for up to 2 seconds for the modal/alert box to appear.

#### Ticket

[CTSKF-615](https://dsdmoj.atlassian.net/browse/CTSKF-615)

#### Why

This should hopefully fix the flickering test in our CI/CD tests:

Failures:

  1) Delete user when manager can index and delete users
     Failure/Error: warning = page.driver.browser.switch_to.alert

     Selenium::WebDriver::Error::NoSuchAlertError:
       no such alert
         (Session info: headless chrome=[redacted])

#### How

Use [Capybara::Session#accept_alert](https://www.rubydoc.info/gems/capybara/Capybara%2FSession:accept_alert) to wait for the alert box for up to 2 seconds.

[CTSKF-615]: https://dsdmoj.atlassian.net/browse/CTSKF-615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/features/users/delete_user_spec.rb
+++ b/spec/features/users/delete_user_spec.rb
@@ -16,7 +16,7 @@ RSpec.feature 'Delete user', :js, type: :feature do
       within row do
         find_link('Delete').click
       end
-
+      sleep 1.second
       warning = page.driver.browser.switch_to.alert
       expect(warning.text).to eql "Are you sure you want to delete #{other_user.name}'s account?"
       warning.accept
--- a/spec/features/users/delete_user_spec.rb
+++ b/spec/features/users/delete_user_spec.rb
@@ -14,12 +14,10 @@ RSpec.feature 'Delete user', :js, type: :feature do
 
       row = page.find(%(tr[data-user-id="#{other_user.id}"]))
       within row do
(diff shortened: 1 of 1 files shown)


spec/features/users/delete_user_spec.rb
```diff
@@ -16,7 +16,7 @@
       within row do
         find_link('Delete').click
       end
-
+      sleep 1.second
```
jrmhaig, line 19: Is it possible to use something like this: https://stackoverflow.com/a/21342673

This is using an `expect` to check for something (in the example, some CSS) on the new page after it has loaded. From some similar Stack Overflow pages it appears that the `have_css` function has a `wait` option to specify a timeout.

Obsiye: thank you, i found accept_alert to suffice after researching your suggestion. It seems to work well and has a wait  option

PR #1882, opened 2024-01-30
Get the Defendant MAAT Reference from the offence_summaries array

Obsiye:
### 

#### What

Get the V2 Defendant **MAAT Reference** from the `defendant_summaries["offence_summaries"]` array

#### Ticket

[board ticket description](link)

#### Why

The true source of maat_reference in the V2 API is from the `defendant_summaries["offence_summaries"]` data not from the `defendant_summaries["representation_order"]`.

#### How

- Add maat_reference method to CDAPI::Defendant model
- Get the maat_reference from the offence_summaries not the representation_order object in the defendant results json
- Update test for the new public method CDAPI::Defendant#maat_reference

### DIFF
--- a/spec/models/cd_api/defendant_spec.rb
+++ b/spec/models/cd_api/defendant_spec.rb
@@ -31,4 +31,23 @@ RSpec.describe CdApi::Defendant, type: :model do
       it { is_expected.to be_falsey }
     end
   end
+
+  describe '#maat_reference' do
+    subject { defendant.maat_reference }
+
+    let(:defendant) { build(:defendant, offence_summaries:) }
+    let(:offence_summaries) { [build(:offence_summary, laa_application:)] }
+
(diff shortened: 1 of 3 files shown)


spec/models/cd_api/defendant_spec.rb
```diff
@@ -31,4 +31,23 @@
...
+  describe '#maat_reference' do
+    subject { defendant.maat_reference }
+
+    let(:defendant) { build(:defendant, offence_summaries:) }
+    let(:offence_summaries) { [build(:offence_summary, laa_application:)] }
+
+    context 'when maat_reference present' do
+      let(:laa_application) { build(:laa_application, reference: '2123456') }
+
+      it { is_expected.to be_truthy }
```
Obsiye, line 44: should be

`is_expected.to eq '2123456'`

PR #2716, opened 2024-12-03
ACD-472: Update Ruby to v. 3.3.6

alexdesi:
#### What
Update Ruby to v. 3.3.6

#### Ticket
https://dsdmoj.atlassian.net/browse/ACD-472

#### Why
To maintain the platform updated, and get all the security patches of ruby lang.


app/models/search.rb
```diff
@@ -6,6 +6,8 @@
 class Search
   include ActiveModel::Model
 
+  attr_accessor :filter, :term, :dob, :version2
+
```
alexdesi, line 10: I moved this on top, to easily check the attributes

app/models/search.rb
```diff
@@ -62,18 +62,13 @@ def query_cd_api
...
-  def defendant_name_query
-    CourtDataAdaptor::Query::Defendant::ByName.new(term, dob:)
+    case filter
+    when 'case_reference'
+      CourtDataAdaptor::Query::ProsecutionCase.new(term)
+    when 'defendant_reference'
+      CourtDataAdaptor::Query::Defendant::ByReference.new(term)
+    when 'defendant_name'
+      CourtDataAdaptor::Query::Defendant::ByName.new(term, dob:)
+    end
```
alexdesi, line 72: This is to simplify `query_cda`. There is no needs of dynamic call `send(:"#{filter}_query")` plus three private methods.
A simple case switch does a great job.

spec/lib/court_data_adaptor/query/defendant/by_name_spec.rb
```diff
@@ -29,7 +29,7 @@ def self.resource
 
     context 'with mocking' do
       let(:resource) { self.class.resource }
-      let(:resultset) { instance_double('ResultSet') }
+      let(:resultset) { instance_double('ResultSet') } # rubocop:disable RSpec/StringAsInstanceDoubleConstant
```
alexdesi, line 32: Rubocos is complaining about this, let's do it later.
This tests also needs refactor, it uses too many stubbed methods.

spec/support/capybara_extensions.rb
```diff
@@ -70,15 +70,15 @@ def has_govuk_error_field?(model, field, error_text = nil)
     end
 
     def has_govuk_detail_summary?(text, options = {})
-      has_selector?(detail_summary_selector, **options.merge(text:))
+      has_selector?(detail_summary_selector, text:, **options)
```
alexdesi, line 73: there is not need to merge text into options.
It's possible to pass it in as argument.

mo-zag (approved):

PR #2820, opened 2025-02-21
ACD-590 - Change email address to new team's address

exonian:
#### What
Changes the email address listed on the contact page, the email address used by the email service, and the email address listed as the maintainer of the Docker image.

#### Ticket

[Change all instances of assess-a-claim email to Access-Court-Data where possible](https://dsdmoj.atlassian.net/browse/ACD-590)

#### Why
The old email address was that of the team that previously owned the service, as opposed to the one that now does.

#### How
The email address is set explicitly in the Docker file. The other two instances are variables used in places throughout the survey.


alexdesi (approved):
It looks good to me,
let's fix the warning later 👍

PR #2821, opened 2025-02-21
Recreate yarn.lock to update JS packages.

alexdesi:
#### What
Update yarn.lock


.circleci/config.yml
```diff
@@ -291,14 +291,22 @@ jobs:
...
       - build-temp-image
-      - snyk/scan:
-          token-variable: SNYK_TOKEN
-          docker-image-name: vcd:scan
-          target-file: ./docker/Dockerfile
-          organization: "legal-aid-agency"
-          severity-threshold: "critical"
-          fail-on-issues: true
-          no-output-timeout: 20m
+      - run:
```
exonian, line 294: Why the change from the orb to a hand-written command? If it's just following a snippet elsewhere, I think it's worth looking at what the actual change is and seeing if we can just incorporate that into what we already have.

Is the point of it to get the `policy-path`?

~Makefile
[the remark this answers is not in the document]
alexdesi: ops, removed!

.circleci/config.yml
alexdesi: it was an experiment, which didn't work. 
`fail-on-issues: false` seems a temporary viable option to me.

alexdesi: I reset that to true, and removed the Cypress lib.

exonian (approved):
CONDITIONAL ON: update the PR description to reflect what we're now doing: no longer installing cypress, and now not ignoring the snyk failure (which now shouldn't happen)

PR #2823, opened 2025-03-10
ACD-594: Disallow indexing

patrick-laa:
#### What
Instruct Google and other crawlers not to index the site in any environment. 

#### Ticket

[Hide VCD from search engines](https://dsdmoj.atlassian.net/browse/ACD-594)

#### Why
Because these pages are not for public consumption and their presence on Google may cause confusion.

"VCD’s intended users have other ways of getting the link (from emails, intranet, bookmarks etc) so the downsides of it being on search engines outweigh any benefits."

#### How
Do this by:
~- Adding a global directive in robots.txt~
- Adding a `meta` tag on each individual page

(robots.txt tells crawlers whether they can crawl pages. SInce Google already has crawled our pages and believes the content is indexable, we want to allow the crawlers to keep crawling so they at some point crawl back and notice that the pages are now `noindex`.)

Note that these directives _ask_ Google not to index us. We cannot _force_ Google not to unless we make it impossible for Googlebot to load any pages on the site.

Finally, note that for me to run bundler, given my processor on my laptop, certain adjustments to the Gemfile are necessary. These aren't directly relevant to the ticket, but I hope they're small enough it's OK to include here rather than a separate PR.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,2 @@
-# See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow: /
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,2 +1,1 @@
-User-agent: *
-Disallow: /
+# See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -190,6 +190,7 @@ GEM
     faraday_middleware (1.2.1)
       faraday (~> 1.0)
     ffi (1.17.0-arm64-darwin)
+    ffi (1.17.0-x86_64-darwin)
     ffi (1.17.0-x86_64-linux-gnu)
     ffi (1.17.0-x86_64-linux-musl)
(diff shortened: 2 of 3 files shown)


public/robots.txt
```diff
@@ -1 +1,2 @@
-# See https://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow: /
```
exonian, line 2: My understanding (albeit entirely from [this google for developers documentation](https://developers.google.com/search/docs/crawling-indexing/robots/intro#what-is-a-robots.txt-file-used-for)) was that not letting the robots in actually increased the risk of the page being indexed, as it could enter the index by being linked to from elsewhere (not a risk for non-prod environments, admittedly) and then the robot wouldn't be able to access the page to get the `noindex` instruction.

Are there search engines which simply don't respect `noindex` (but that do respect `robots.txt`, such that we simply have to keep their bots out? If so, could our robots.txt instruction be more specific?

I am happy to defer on this, I'm just conscious I read a thing so would like to know where we're superceding that

exonian (dismissed):

exonian (approved):

PR #2849, opened 2025-03-13
ACD-611: Remove PII from CDAPI request logs

patrick-laa:
#### What
Users of VCD conduct searches of personally identifiable information (PII), including the names,  dates of birth and NI numbers of defendants. 

At present, when a search is conducted using CDAPI, VCD will put search terms in the logs that get sent to OpenSearch, meaning that OpenSearch contains, amongst its logs, a bunch of PII that it shouldn’t. 

#### Ticket

[Remove PII from logs in VCD and CDAPI](https://dsdmoj.atlassian.net/browse/ACD-611)

#### Why
We have responsibilities governing the storage and security of that information, so need to limit the number of places that information is stored.

#### How
Intercept log messages generated by ActiveResource, parse the request URI, and filter out any query params that are blacklisted Rails filter parameters (ensuring the appropriate terms are indeed blacklisted)


spec/lib/active_resource/log_subscriber_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_reader :output
+
+  def request_without_filtered_logging(input)
+    @output = input
+  end
+end
+
+RSpec.describe ActiveResource::LogSubscriber do
+  subject(:filtered_log) do
+    ActiveResourceLogSubscriberTestImplementation.new.tap { _1.request_with_filtered_logging(event) }.output
```
alexdesi, line 11: I am not a fan of `_1`, not very readable, but it's ok for now.
Ruby 3.4 introduces the `it` block param (like kotlin), which is pretty cool :)

config/initializers/active_resource_filtered_logging.rb
```diff
@@ -0,0 +1,39 @@
...
+# This monkey-patch exists because CdApi uses ActiveResource which, by default, logs
+# all query params of outgoing GET requests. However, our requests include PII that we
+# don't want ending up in the logs (e.g. defendant date of birth). So we need to intercept
+# the log commands and scrub out sensitive query params
+# (based on a gist found at https://gist.github.com/vraravam/1319094)
+module ActiveResource
+  class LogSubscriber < ActiveSupport::LogSubscriber
+    def request_with_filtered_logging(event)
+      request_uri = event.payload[:request_uri]
+      start_of_query_string = request_uri.index("?")
```
alexdesi, line 13: it seems you don't need the idex,  `include?` sounds a better option here

alexdesi (approved):
it's great you added tests,  
it looks good! ✅

PR #2853, opened 2025-03-14
ACD-255: Sorting hearing counsels

patrick-laa:
#### What
When using the v2 endpoints/CDAPI, whenever the `/prosecution_cases/<id>` screen is accessed, `TableSorters::HearingsProviderSorter` sorts the hearings by the defence_counsel_list property. `HearingSummaryDecorator#defence_counsel_list` calls a method called `attended_defence_counsels` which assumes that the hearing summary has a `day` attribute set. _HOWEVER_, hearings do not have a "day" by default (because hearings can run over multiple days). The `day` attribute is only set when the hearing summary object is being used to display information about a specific day of the many which can be associated with a hearing (e.g. `CaseSummaryDecorator#sorted_hearing_summaries_with_day`). At the point of sorting, `day` has not been set, so `day.strftime` in `attended_defence_counsels` errors out.

This has happened 1.4k times in the last 30 days according to Sentry, and basically happens any time there are defence counsels associated with any of the hearings attached to a case.

Note that it's impossible to repro this locally without modifying our HMCTS mock because the mock hearing summary payload doesn't include the `defenceCounsel` attribute.

### DIFF
--- a/app/decorators/cd_api/hearing_summary_decorator.rb
+++ b/app/decorators/cd_api/hearing_summary_decorator.rb
@@ -44,8 +44,15 @@ module CdApi
     end
 
     def attended_defence_counsels
+      # Sometimes a hearing summary pertains to a specific day of a hearing.
+      # (c.f. CdApi::CaseSummaryDecorator.sorted_hearing_summaries_with_day)
+      # In those cases, we are only interested in the defence counsels who
+      # attended on that specific day. Otherwise, we are interested in
+      # all defence counsels
+      return defence_counsels unless day
+
       defence_counsels.select do |defence_counsel|
-        defence_counsel.attendance_days.include?(day.strftime('%Y-%m-%d'))
+        defence_counsel.attendance_days&.include?(day.strftime('%Y-%m-%d'))
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/decorators/cd_api/hearing_summary_decorator.rb
```diff
@@ -44,8 +44,15 @@ def mapped_defence_counsels
...
+      # Sometimes a hearing summary pertains to a specific day of a hearing.
+      # (c.f. CdApi::CaseSummaryDecorator.sorted_hearing_summaries_with_day)
+      # In those cases, we are only interested in the defence counsels who
+      # attended on that specific day. Otherwise, we are interested in
+      # all defence counsels
+      return defence_counsels unless day
+
       defence_counsels.select do |defence_counsel|
-        defence_counsel.attendance_days.include?(day.strftime('%Y-%m-%d'))
+        defence_counsel.attendance_days&.include?(day.strftime('%Y-%m-%d'))
```
alexdesi, line 55: good spot,
it'd be great if you can make it a bit more explicit, for example:
```
  defence_counsels.
  .reject { |dc| dc.attendance_days.nil? } # TODO:add an explanation/note here
  .select do |defence_counsel|
    defence_counsel.attendance_days&.include?(day.strftime('%Y-%m-%d'))
  end
```

alexdesi (dismissed):
✅

alexdesi (approved):
✅

PR #2854, opened 2025-03-14
BUG: Fix typo

patrick-laa:
On UAT, when trying to establish if a defendant is linked by MAAT ID when showing the "raw data" in the UI, we hit an error looking for "offence_summaries" when we've only defined "offence_summary". Note that this doesn't show up in the test suite because our FactoryBot factory plus ActiveResource combine to make our test fixtures actually have an "offence_summaries" attribute even though it's not defined on the class.

### DIFF
--- a/app/models/cd_api/defendant.rb
+++ b/app/models/cd_api/defendant.rb
@@ -2,7 +2,7 @@
 
 module CdApi
   class Defendant < BaseModel
-    has_many :offence_summary, class_name: 'cd_api/offence_summary'
+    has_many :offence_summaries, class_name: 'cd_api/offence_summary'
 
     def linked?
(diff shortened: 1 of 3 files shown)


app/models/cd_api/defendant.rb
```diff
@@ -2,7 +2,7 @@
 
 module CdApi
   class Defendant < BaseModel
-    has_many :offence_summary, class_name: 'cd_api/offence_summary'
+    has_many :offence_summaries, class_name: 'cd_api/offence_summary'
```
alexdesi, line 5: great spot, 
what about adding a small test in the  
`models/cd_api/defendant_spec.rb` ?

PR #2858, opened 2025-03-17
BUG: Handle nil offence summaries

patrick-laa:
#### What
When `GET <cd api>/v2/offence_summaries?defendant_id=<id>` returns a 404 (common on UAT), `CdApi::Defendant#offence_summaries` is nil, so `#maat_references` raises an error.

This is related to the previous bug where `offence_summaries` didn't exist at all, as it was only defined on the model as `offence_summary`. I should have checked more UAT cases before merging.


alexdesi (approved):
it looks good ✅

PR #2870, opened 2025-03-19
Switch to dependabot weekly grouped updates

patrick-laa:
#### What
Move dependabot updates to a weekly cadence, and have them grouped by dependency type

#### Why
We have 30 open dependabot PRs. They're not getting tested or merged, so our dependencies are falling further and further behind, and they're generating loads of noise in email alerts etc.


alexdesi (approved):
💯

PR #2877, opened 2025-03-25
ACD-623: Staging should point to CDA test

patrick-laa:
Note that this will need some secrets updating to accompanying, but that's not done via PR


alexdesi (approved):
✅

PR #2885, opened 2025-04-03
Remove "defendants_search" feature flag

alexdesi:
Ticket: https://dsdmoj.atlassian.net/browse/ACD-640

#### What
Remove "defendants_search" feature flag

#### Why
- It is not used
- It is always set to true for all the environments
- To clean the code
- To run test faster


app/controllers/searches_controller.rb
```diff
@@ -12,7 +12,7 @@ def new
   end
 
   def create
-    @search = Search.new(filter:, term:, dob:, version2: version_2?)
+    @search = Search.new(filter:, term:, dob:, version2: true)
```
alexdesi, line 15: at the end of removing all feature flags,
`version2: true` arg can be removed.

spec/features/breadcrumbs_spec.rb
```diff
@@ -51,7 +51,7 @@
       end
     end
 
-    context 'when on case details page' do
+    context 'when on case details page', :stub_defendants_case_search do
```
alexdesi, line 54: before, the breadcrumbs, was tested against V1 API endpoint.
Not it uses the V2, which needs `stub_defendants_case_search`

alexdesi:
All those *_spec.rb files tests the old V1 CDA, wich is not used anymore,
so I'll removed them.
(more is coming in the next PRs)

app/controllers/searches_controller.rb
patrick-laa: Ah, it's because the `hearing` flag also sometimes is used to decide whether to use `version2` in Search?

patrick-laa (approved):
932 LOC deleted! Love it!

app/controllers/searches_controller.rb
alexdesi: Yes,
`DEFENDANTS_SEARCH_CDAPI_V2` might have been a better name than `DEFENDANTS_SEARCH.`

PR #2890, opened 2025-04-10
ACD-638 and 639: Remove the hearing summaries feature flag

patrick-laa:
#### What
We will never switch the hearing summaries feature flag off again, so this PR removes it. 90% of the ticket is removing or refactoring specs that assume the feature flag is switched off.

#### Ticket

[ACD-639](https://dsdmoj.atlassian.net/browse/ACD-639)

[ACD-639]: https://dsdmoj.atlassian.net/browse/ACD-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/prosecution_cases/show.html.haml
```diff
@@ -1,15 +1,3 @@
...
-      = render partial: 'cracked_ineffective_trial', locals: { prosecution_case: @prosecution_case }
-
-      = render partial: 'defendants', locals: { prosecution_case: @prosecution_case }
-
-      = render partial: 'hearing_summaries', locals: { prosecution_case: @prosecution_case }
-
-      - if Rails.configuration.x.display_raw_responses
-        = render partial: 'raw_response', locals: { prosecution_case: @prosecution_case }
\ No newline at end of file
+= render partial: 'prosecution_cases/cd_api/show_v2', locals: { case_summary: @prosecution_case }
```
alexdesi, line 3: it would be nice to expand the `prosecution_cases/cd_api/show_v2` inside the show view,
since it is the only place it is used.
So, we can remove that partial.

We can the remove the test `prosecution_cases/cd_api/_show_v2.html.haml_spec.rb`, a quick look suggests it does not seems to provide much value.

spec/spec_helper.rb
```diff
@@ -60,7 +60,7 @@
   # Allows RSpec to persist some state between runs in order to support
   # the `--only-failures` and `--next-failure` CLI options. We recommend
   # you configure your source control system to ignore this file.
-  # config.example_status_persistence_file_path = "spec/examples.txt"
+  config.example_status_persistence_file_path = "spec/examples.txt"
```
alexdesi, line 63: it sounds good 👍

spec/views/prosecution_cases/show.html.haml_spec.rb
```diff
@@ -1,23 +1,28 @@
 # frozen_string_literal: true
 
-RSpec.describe 'prosecution_cases/show', type: :view do
-  subject(:render_view) { render }
+RSpec.describe 'prosecution_cases/cd_api/_show_v2.html.haml', type: :view do
```
alexdesi, line 3: as suggested before, we could get rid off `cd_api/_show_v2.html.haml`,
and maybe the relevant spec.

alexdesi (commented):
it looks good,
the main suggestion here is to remove the partial `cd_api/_show_v2.html.haml`,
if you are busy tell me, we can do it afterwards.

alexdesi (approved):
thanks for removing `_show_v2` 🚀 ✅

PR #2896, opened 2025-04-14
ACD-539: Hearing details pagination and error handling

patrick-laa:
#### What
- Load all hearing details needed to display a hearing details page for a court application
- Add next/previous functionality
- Handle most obvious failure cases gracefully
-
#### Ticket

[As a case worker, I want to access hearing details](https://dsdmoj.atlassian.net/browse/ACD-539)

### DIFF
--- a/app/controllers/hearing_days_controller.rb
+++ b/app/controllers/hearing_days_controller.rb
@@ -1,0 +1,65 @@
+class HearingDaysController < ApplicationController
+  before_action :load_and_authorize_application, :load_hearing_days, :load_hearing_details,
+                :load_hearing_events
+
+  add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
+  add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
+
+  def show
+    add_breadcrumb @application.application_title
+    add_breadcrumb t(".breadcrumb", day: @hearing_day.day_string)
+  end
+
+  private
+
+  def load_and_authorize_application
+    @application = CourtDataAdaptor::Query::CourtApplication.new(params[:court_application_id]).call
+    authorize! :show, @application
+  rescue JsonApiClient::Errors::ServiceUnavailable => e
+    Sentry.capture_exception(e)
+    redirect_to controller: :errors, action: :internal_error
+  rescue JsonApiClient::Errors::NotFound
+    redirect_to controller: :errors, action: :not_found
+  end
+
+  def load_hearing_days
+    load_hearing_day
+    load_adjacent_hearing_days
+  end
+
+  def load_hearing_details
+    details = CdApi::Hearing.find(@hearing_day.hearing.id, params: { date: @hearing_day.date })
+    @hearing_details = helpers.decorate(details, CdApi::HearingDecorator)
+  rescue ActiveResource::ServerError, ActiveResource::ClientError => e
+    logger.error e
+    Sentry.capture_exception(e)
+    redirect_to controller: :errors, action: :internal_error
+  end
+
+  def load_hearing_events
+    @hearing_events = CdApi::HearingEvents.find(@hearing_day.hearing.id, { date: @hearing_day.date })
+  rescue ActiveResource::ResourceNotFound
+    logger.warn "No hearing events found for hearing #{params[:hearing_id]}"
+  rescue ActiveResource::ServerError, ActiveResource::ClientError => e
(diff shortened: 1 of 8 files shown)


app/controllers/hearing_days_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def show
+    add_breadcrumb @application.application_title
+    add_breadcrumb t(".breadcrumb", day: @hearing_day.day_string)
+  end
+
+  private
+
+  def load_and_authorize_application
+    @application = CourtDataAdaptor::Query::CourtApplication.new(params[:court_application_id]).call
```
muthusubra, line 16: Should we add parameter validation to ensure parameters are present?

app/controllers/hearing_days_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def load_adjacent_hearing_days
+    current_day_index = flat_hearing_days.index(@hearing_day)
+
+    @next_day = flat_hearing_days[current_day_index + 1]
+    @previous_day = flat_hearing_days[current_day_index - 1] if current_day_index.positive?
+  end
+
+  def flat_hearing_days
+    @flat_hearing_days ||= @application.hearing_days_sorted_by("asc")
```
muthusubra, line 65: Can use a constant for sort direction

patrick-laa: Good idea! Done.

app/controllers/hearing_days_controller.rb
patrick-laa: A nice Rails thing helps us out here - because the params here are in the path rather than the querystring, they have to be present for the route matcher to work, so if they're missing this method wouldn't be called in the first place - it'd be handled as a 404 automatically.

app/decorators/cd_api/hearing_decorator.rb
```diff
@@ -21,6 +21,12 @@ def prosecution_counsels_list
       safe_join(formatted_prosecution_counsels, tag.br)
     end
 
+    def judiciary_list
+      return t('generic.not_available') if hearing.judiciary.none?
+
+      safe_join(hearing.judiciary.map { |jd| "#{jd.title} #{jd.first_name} #{jd.last_name}" }, tag.br)
```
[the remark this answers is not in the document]
patrick-laa: Annoyingly it does seem to be wildly inconsistent in the existing codebase. `DefenceCounselsListService` uses first name, middle name, last name. `HearingDecorator#prosecution_counsels_list` uses just first name and last name. When the `attendees_v2` partial constructs judge names it uses title, first name and last name. I don't know enough to know whether this is intentional and important.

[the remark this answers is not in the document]
patrick-laa: I've done a small refactor so that for judiciary a least, both the prosecution hearings page and the appeal hearings page use the same formatting code to avoid further divergence.

app/controllers/hearing_days_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  def load_hearing_days
+    load_hearing_day
+    load_adjacent_hearing_days
+  end
+
+  def load_hearing_details
+    details = CdApi::Hearing.find(@hearing_day.hearing.id, params: { date: @hearing_day.date })
+    @hearing_details = helpers.decorate(details, CdApi::HearingDecorator)
+    @hearing_details.current_sitting_day = @hearing_day.sitting_day
+    @hearing_details.skip_mapping_counsels_to_defendants = true
```
muthusubra, line 36: What is the purpose of this flag? Skip the mapping as it isn't needed?

patrick-laa: So the logic that works out which defence counsels attended on a given day also tries to work out which defendants were being represented by which counsel (so it can say things like "Barry McBarristerface (for Deffy McDefendant)" in the UI. The logic is both broken and redundant for appeal hearings (broken because it specifically tries to pull defendants from a prosecution case object that isn't available, and redundant because appeals don't concern themselves with multiple defendants), but rather than re-implement the "which counsels were there on which day" logic, I wanted to reuse the existing logic, simply skipping over the broken/redundant stuff, so that we end up with "Barry McBarristerface" and no parentheses.

muthusubra: Okay, got it. Thank you!

PR #2897, opened 2025-04-14
ACD-522 Show user appeal court application link within the search results

mo-zag:
#### What
This ticket is about adding a court application link to kick off the user journey for new appeals.
#### Ticket

[ACD-522](https://dsdmoj.atlassian.net/browse/ACD-522)

#### Why
So that the user can start the appeals journey by clicking the provided link.

#### How
When the search page is run, the VCD API responds with a defendant_summaries payload, which includes an embedded application_summaries payload.

### DIFF
--- a/app/views/results/_defendant_appeals.html.haml
+++ b/app/views/results/_defendant_appeals.html.haml
@@ -1,0 +1,27 @@
+%table.govuk-table
+  %caption.govuk-table__caption.govuk-visually-hidden
+    = t('search.result.caption')
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.name')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.date_of_birth')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.prosecution_case_reference')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.arrest_summons_number')
+      %th.govuk-table__header{ scope: 'col' }= t('search.result.defendant.court_application_title')
+  %tbody.govuk-table__body
+    - results.each do |defendant|
+      %tr.govuk-table__row
+        %td.govuk-table__cell
+          = link_to defendant.name, defendant_link_path(defendant, defendant.prosecution_case_reference), class: 'govuk-link'
+        %td.govuk-table__cell
+          = l(defendant.date_of_birth&.to_date)
+        %td.govuk-table__cell
+          = link_to defendant.prosecution_case_reference, prosecution_case_path(defendant.prosecution_case_reference), class: 'govuk-link'
+
+        %td.govuk-table__cell
+          = defendant.arrest_summons_number
+        %td.govuk-table__cell
+          - if defendant&.application_summaries&.present?
+            = render partial: 'results/defendant_court_applications', locals: { results: defendant.application_summaries }
+          - else 
+            = t('search.result.defendant.no_court_application_title')
--- a/app/views/results/_defendant_court_applications.html.haml
+++ b/app/views/results/_defendant_court_applications.html.haml
@@ -1,0 +1,4 @@
+- results.each do |defendant_court_applications|
+  
+  -# = link_to defendant_court_applications.title, court_application_path(defendant_court_applications.id), class: 'govuk-link'
+  = link_to defendant_court_applications.title, "#{'#' + defendant_court_applications.id}", class: 'govuk-link'
--- a/app/views/searches/_results.html.haml
+++ b/app/views/searches/_results.html.haml
@@ -1,8 +1,10 @@
 = render partial: 'results_header', locals: { term: term, dob: dob }
-
 - if results.empty?
   = render 'results/none'
 - elsif results.first.is_a? CourtDataAdaptor::Resource::ProsecutionCase
   = render partial: 'results/prosecution_case', locals: { results: results }
 - else
-  = render partial: 'results/defendant', locals: { cookies_preferences_set: cookies_preferences_set, results: results }
+  - if ENV['SHOW_APPEALS'].to_s.downcase == 'true'
+    = render partial: 'results/defendant_appeals', locals: { cookies_preferences_set: cookies_preferences_set, results: results }
+  - else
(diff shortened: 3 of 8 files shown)


app/views/searches/_results.html.haml
```diff
@@ -1,8 +1,10 @@
 = render partial: 'results_header', locals: { term: term, dob: dob }
-
 - if results.empty?
   = render 'results/none'
 - elsif results.first.is_a? CourtDataAdaptor::Resource::ProsecutionCase
   = render partial: 'results/prosecution_case', locals: { results: results }
 - else
-  = render partial: 'results/defendant', locals: { cookies_preferences_set: cookies_preferences_set, results: results }
+  - if ENV['SHOW_APPEALS'].to_s.downcase == 'true'
```
patrick-laa, line 7: We've got a utility class for this that we might as well use: `FeatureFlag.enabled?(:show_appeals)`.

Also, probably worth setting the flag in at least the dev `deployment.yaml` file

config/locales/en/views/searches.yml
```diff
@@ -20,11 +20,13 @@ en:
       defendant:
         arrest_summons_number: ASN
         caption: Defendants
+        court_application_title: Related Applications
```
patrick-laa, line 23: Likewise I don't believe this should have a capital A.

app/views/results/_defendant_court_applications.html.haml
```diff
@@ -0,0 +1,3 @@
+- results.each do |defendant_court_applications|
+  -# = link_to defendant_court_applications.title, court_application_path(defendant_court_applications.id), class: 'govuk-link'
+  = link_to defendant_court_applications.title, "#{'#' + defendant_court_applications.id}", class: 'govuk-link'
```
patrick-laa, line 3: Just eyeballing it, but it looks to me like we'd probably want to add a little margin-bottom here, otherwise it's hard to visually distinguish two appeals as being distinct links:
Screenshot 2025-04-15 at 15 04 40

app/views/results/_defendant_appeals.html.haml
```diff
@@ -0,0 +1,27 @@
...
+          = link_to defendant.name, defendant_link_path(defendant, defendant.prosecution_case_reference), class: 'govuk-link'
+        %td.govuk-table__cell
+          = l(defendant.date_of_birth&.to_date)
+        %td.govuk-table__cell
+          = link_to defendant.prosecution_case_reference, prosecution_case_path(defendant.prosecution_case_reference), class: 'govuk-link'
+
+        %td.govuk-table__cell
+          = defendant.arrest_summons_number
+        %td.govuk-table__cell
+          - if defendant&.application_summaries&.present?
```
patrick-laa, line 24: It would be good to have a little test coverage here. Maybe a feature spec that uses VCR to pull in a result with a couple of court applications, and assert that when the feature flag is switched on their titles are visible?

app/views/results/_defendant_appeals.html.haml
```diff
@@ -0,0 +1,27 @@
+%table.govuk-table
```
patrick-laa, line 1: Most of this is copy-pasted from an existing view. Would it make sense to just have one view, and use the feature flag to toggle whether we display the final `th`/`td`? Just to avoid duplication?

mo-zag: I think its easier and cleaner this way for now once we remove the faeture flag this can be removed.

app/views/results/_defendant_court_applications.html.haml
mo-zag: I'll add the margin this will change if you have more than two application it does not work.

app/views/searches/_results.html.haml
mo-zag: Yes, we do — but I wasn’t sure if this utility class might be removed in the future, since we're aiming to phase out some features. I kept it simple so it can be easily removed in one place if needed.

patrick-laa: Ah, yeh, you mentioned this before. So we're looking to remove some specific old feature flags, but we're very much _not_ trying to remove the nice clean feature flag mechanism we have. It makes testing a lot easier if we can simply do an `allow(FeatureFlag).to receive(:enabled?).with(:show_appeals).and_return ...` rather than have to start messing with `ENV` in our tests.

patrick-laa: (But if you'd prefer to work directly with ENV then fair enough - it still delivers the right result)

config/locales/en/views/searches.yml
```diff
@@ -20,11 +20,13 @@ en:
...
         date_of_birth: Date of birth
         heading: Defendants
         maat_number: MAAT number
         name: Name
         national_insurance_number: NI number
+        no_court_application_title: No Related Application
```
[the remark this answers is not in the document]
mo-zag: should be done now

app/views/searches/_results.html.haml
```diff
@@ -1,8 +1,10 @@
 = render partial: 'results_header', locals: { term: term, dob: dob }
-
 - if results.empty?
   = render 'results/none'
 - elsif results.first.is_a? CourtDataAdaptor::Resource::ProsecutionCase
   = render partial: 'results/prosecution_case', locals: { results: results }
 - else
-  = render partial: 'results/defendant', locals: { cookies_preferences_set: cookies_preferences_set, results: results }
+  - if FeatureFlag.enabled?(:show_appeals)
+    = render partial: 'results/defendant_appeals', locals: { cookies_preferences_set: cookies_preferences_set, results: results }
```
patrick-laa, line 8: Do you really need to pass `cookies_pereferences_set` here? It looks to me like the old view didn't actually need it, it was just a copy-paste mistake

patrick-laa (approved):

spec/support/webmock/court_data_api/webmock.rb
```diff
@@ -62,6 +62,16 @@
     )
   end
 
+  config.before(:each, :stub_appeals_defendants_case_search) do
```
patrick-laa, line 65: You don't actually need this. If you delete it, and the reference to it in the spec file, the specs still pass fine, because the `describe` block at the top of the file already has the `stub_defendants_case_search` which is identical to this.

PR #2898, opened 2025-04-15
ACD-567 & 569-b: Show appeal results

patrick-laa:
#### What
Use the title and result code of an application to show a result string if possible on the application summary and hearing details pages

#### Ticket

[As a caseworker I want to know the result of an appeal](https://dsdmoj.atlassian.net/browse/ACD-567)
[As a caseworker I want to view hearing details](https://dsdmoj.atlassian.net/browse/ACD-569)


app/views/hearing_days/show.html.haml
```diff
@@ -82,3 +82,9 @@
         = t('.event.none')
 
     = render partial: 'hearings/court_applications_v2', locals: { hearing: @hearing_details.hearing }
+
+    - if @application.subject_summary.proceedings_concluded
```
mo-zag, line 86: Is there a case for changing this to @application&.subject_summary&.proceedings_concluded to make it safer and avoid potential nil errors?

patrick-laa: I dunno, if `@application` is nil here, or if `@application.subject_summary` doesn't return a SubjectSummary, then that's a sign something serious has gone wrong (and in practice this view would have blown up 40 lines higher up). To me that feels like a "this page is broken and we shouldn't try to  recover and show what would definitely be misleading data" kind of scenario.

config/i18n-tasks.yml
```diff
@@ -120,6 +120,7 @@ ignore_unused:
 - date.formats.*
 - activemodel.errors.models.*
 - activerecord.errors.models.*
+- court_applications.results.*
```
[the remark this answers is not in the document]
patrick-laa: Annoyingly `i18n-tasks` isn't as clever as it thinks it is, and it false positives on the new keys a "missing" because it can't see where they get used in the code.

mo-zag (approved):
Looks ok to me just one comment if you think its ok then all good from me.

PR #2917, opened 2025-05-06
ACD-688, 689, 690: Ephemeral deployments

patrick-laa:
#### What
Switch from manual k8s scripts to using helm. This will make it easier to do ephemeral deployments in the dev environment as a next step. Note, before this can be merged #2918 needs to be merged.


alexdesi (approved):
✅

PR #2925, opened 2025-05-08
ACD-662: Remove never-switched-on feature flag

patrick-laa:
#### What
The `defendants_page` FF cannot be switched on because uses the v2 defendants endpoint in CDA which doesn't include the plea history. We will eventually switch to that endpoint once we have new designs to surface the plea history in other ways. And when we get to that, CDAPI will be gone, so we will have to re-implement this anyway.

So for now we clear out the old unused code as in its current form it can never be used.

#### Ticket

[[VCD] Remove Feature Flag "defendants_page"](https://dsdmoj.atlassian.net/browse/ACD-662)


config/locales/en/views/searches.yml
```diff
@@ -41,7 +41,6 @@ en:
         providers: Providers attending
       offence:
         caption: Offence list
-        date: Date
```
alexdesi, line 44: just to check, was this removed on purpose?

alexdesi (approved):
this cleaning up makes me so happy 💯 !
🧹🧹🧹

config/locales/en/views/searches.yml
patrick-laa: Yes - offences_v2 had an extra column showing the _date_ of the offence, which isn't available in the v1 API. (Side note: we should ticket up _adding_ the date of the offence to the v1 payload so that we can add this column to the VCD UI without having to switch to v2)

PR #2930, opened 2025-05-09
ACD-678: Load hearing events directly from CDA

patrick-laa:
#### What
As a step towards retiring CDAPI

#### Ticket

[[Discontinue CDAPI] Retire GET /v2/hearings/<uuid>/hearing_events](https://dsdmoj.atlassian.net/browse/ACD-678)

### DIFF
--- a/app/models/cda/base_model.rb
+++ b/app/models/cda/base_model.rb
@@ -1,0 +1,12 @@
+module Cda
+  class BaseModel < ActiveResource::Base
+    self.site = CourtDataAdaptor::Resource::V2.api_url
+
+    def self.headers
+      {
+        'Content-Type' => 'application/json',
+        'Authorization' => "Bearer #{CourtDataAdaptor::Client.new.bearer_token}"
+      }
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/models/cda/base_model.rb
```diff
@@ -0,0 +1,12 @@
+module Cda
+  class BaseModel < ActiveResource::Base
+    self.site = CourtDataAdaptor::Resource::V2.api_url
```
alexdesi, line 3: I've used the same value for `self.site`, but I'd like to change it.
I think we should not use the the `CourtDataAdaptor::Resource::V2.api_url` : `Cda::BaseModel` help to create resources,  and we are using the site url of a resource that we are deprecating.

I think we should not reuse code in `CourtDataAdaptor::Resource`

app/models/cda/hearing_event_log.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Cda
+  class HearingEventLog < BaseModel
+    def self.load(hearing_id, date)
```
alexdesi, line 5: for clarity, what about renaming `load` to  `find_by_hearing_and_date` ?

alexdesi (approved):
it looks good ✅ 
Just a comment about renaming `load`, which sounds generic.

PR #2932, opened 2025-05-09
ACD-675: Load searches directly from CDA

patrick-laa:
#### What
All searches now happen directly against CDA meaning we no longer interact with CDAPI /v2/defendants at all.

#### Ticket

[[Discontinue CDAPI] Retire GET /v2/defendants](https://dsdmoj.atlassian.net/browse/ACD-675)


spec/fixtures/stubs/cda/cases_body.json
```diff
@@ -0,0 +1,339 @@
+{
```
alexdesi, line 1: should we call it with something that reminds the api call?
I called the response for the hearing_results this way `cda/internal_v2_hearing_results_response.json`

app/models/search.rb
```diff
@@ -44,6 +44,6 @@ def self.sanitize_html(html_string)
             if: proc { |search| search.filter.eql?('defendant_name') }
 
   def execute
-    CdApi::SearchService.call(filter:, term:, dob:)
+    CourtDataAdaptor::DefendantSearchService.call(filter:, term:, dob:)
```
alexdesi, line 47: `DefendantSearchService` is a much more clear name! 👍

PR #2934, opened 2025-05-12
ACD-701: Show offence dates

patrick-laa:
#### What
Implement the offence date UI that we designed for the v2 view, now that it's available in the v1 endpoint

#### Ticket

[Surface offence summary `startDate` in VCD](https://dsdmoj.atlassian.net/browse/ACD-701)

#### How

--------

#### TODO (wip)

 - [ ] item 1
 - [X] item 2
 - [X] Has the Acceptance Criteria been met

### DIFF
--- a/spec/views/defendants/_offences.html.haml_spec.rb
+++ b/spec/views/defendants/_offences.html.haml_spec.rb
@@ -11,6 +11,24 @@ RSpec.describe 'defendants/_offences.html.haml', type: :view do
       assign(:defendant, defendant)
     end
 
+    context 'when offence has a start date' do
+      before { allow(offence).to receive(:start_date).and_return('2023-01-01') }
+
+      it 'displays offence start date' do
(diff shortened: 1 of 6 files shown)


spec/views/defendants/_offences.html.haml_spec.rb
```diff
@@ -11,6 +11,24 @@
...
+                                text: %r{01/01/2023})
+      end
+    end
+
+    context 'when offence has no start date' do
+      before { allow(offence).to receive(:start_date).and_return(nil) }
+
+      it 'displays no offence start date' do
+        is_expected.to have_css('.govuk-table__cell',
+                                text: /Not available/)
```
alexdesi, line 28: "Not available" string may be visible in other part of the UI,
in that case this check is too generic.

(but I am not sure how easily we can identify that part of the html)

alexdesi (approved):
I looks good, I've just a concern about a test.

PR #2938, opened 2025-05-14
BUG: Missing attributes in production payloads causing NoMethodErrors

patrick-laa:


app/models/cda/defendant.rb
```diff
@@ -1,7 +1,21 @@
 module Cda
   class Defendant < BaseModel
     def name
-      [first_name, middle_name, last_name].filter_map(&:presence).join(" ")
+      [first_name, try(:middle_name), last_name].filter_map(&:presence).join(" ")
```
alexdesi, line 4: are you sure `try(:middle_name)` is needed?
the &:presence should filter it presence, correct?

alexdesi (approved):
it looks good ✅ 
I am just not sure if `try(:middle_name) is needed.

app/models/cda/defendant.rb
patrick-laa: As discussed, middle_name is not nil, it's not implemented

PR #2950, opened 2025-05-22
ACD-254: Improve handling of invalid dates

patrick-laa:
#### What
Currently if a caseworker enters an invalid date we will re-render the form with an error message but we will throw away what they entered, so they won't be able to see what they did wrong. This PR fixes that by retaining the original values provided in the event of a re-render.


app/models/date_field_collection.rb
```diff
@@ -0,0 +1,33 @@
+# The goal of the class is to ensure that, if the user enters invalid values, like 35 January 2023,
+# we preserve what the user entered so we can show it back to them with an appropriate error message.
+# If we were to directly parse to a Date, invalid values would be rejected and at best we could
+# represent the entire attribute as `nil`, losing the specifics of what was invalid.
+class DateFieldCollection
+  def initialize(params, attribute_name)
```
alexdesi, line 6: it's all right here,
I just wonder if we need the `attribute_name` since :dob is the only one.

alexdesi (approved):
I looks good, the only note is if we could simplify `DateFieldCollection` since DOB is the only date attribute to validate.
✅

PR #2963, opened 2025-05-29
Allow explicit use of turbo

patrick-laa:
#### What
I turned off Turbo Forms to stop form submissions being intercepted (and occasionally broken) by Turbo. BUT in doing so I turned off Turbo's "sneakily turn this link into a button" functionality, which is how sign-out and delete-user links work.

By switching the mode to "opt-in", Turbo won't kick in unless it's explicitly asked via thing like the `turbo_method` annotation. That way it won't interfere with forms, but it will do "sneakily turn this link into a button".


alexdesi (approved):
I looks good! ✅

PR #2964, opened 2025-06-02
ACD-368: Add MAAT reference when unlinking court applications

patrick-laa:
#### What
As a defence against weird situations where the CDA database ends up with multiple LaaReference records for the same subject ID with different MAAT references, when unlinking court applications, pass in the MAAT reference we want to unlink _from_. (We already do this for prosecution cases).

### DIFF
--- a/spec/features/unlink_court_applications_spec.rb
+++ b/spec/features/unlink_court_applications_spec.rb
@@ -1,5 +1,5 @@
 RSpec.feature 'Unlink court applications' do
-  let(:user) { create(:user) }
+  let(:user) { create(:user, username: "kova-a81") }
   let(:unlinked_court_application_id) { 'c07d4116-0d06-4150-b6a4-e412f556d931' }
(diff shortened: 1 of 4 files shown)


spec/features/unlink_court_applications_spec.rb
```diff
@@ -36,6 +36,26 @@
...
+      unlink_stub = stub_request(:patch, /.*court_application_laa_references.*/).with(
+        body: {
+          laa_reference: {
+            subject_id: "6c3eded6-a6d6-4156-940d-e3b5f02deb96",
+            user_name: "kova-a81",
+            unlink_reason_code: 4,
+            unlink_other_reason_text: nil,
+            maat_reference: "1234568"
+          }
+        }.to_json
```
alexdesi, line 49: this test is already using VCR,
my concern is that it'd be even more difficult to maintain if we start mixing VCR and stub_request 🤔

alexdesi (approved):
Thanks for removing the webmock! 
✅

PR #2975, opened 2025-06-06
Bug linked link

patrick-laa:
#### What
On the search results screen we don't have proper code checking if a defendant is linked or not - we always return nil. By wrapping the defendant in a decorator that has a proper `linked?` method, and granting access to the offence_summary->laa_application attributes as objects, we can use our existing "is this linked?" logic.


patrick-laa:
> In my tests, the error was when I was trying to Link the appellant, the defendant was actually showing the error message.

@alexdesi this PR is for a different issue.

PR #2978, opened 2025-06-09
ACD-806: Show error codes

patrick-laa:
When CDA returns a failure response, try to extract an error code from the body, and if one is found and we know a user-friendly string that is appropriate to it, display that string in the UI.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-806)

### DIFF
--- a/lib/court_data_adaptor/errors.rb
+++ b/lib/court_data_adaptor/errors.rb
@@ -43,7 +43,8 @@ module CourtDataAdaptor
         return unless error_response['error_codes']
 
         error_response['error_codes'].filter_map { |code| build_message(code) }.join(" ").presence
-      rescue StandardError
+      rescue StandardError => e
+        Rails.logger.error "CourtDataAdaptor::Errors::ErrorCodeParser error: #{e.message}"
         nil
       end
 
--- a/lib/court_data_adaptor/errors.rb
+++ b/lib/court_data_adaptor/errors.rb
@@ -7,19 +7,22 @@ module CourtDataAdaptor
         @response = response
         @status = response.status
         @errors = response.body
-        @error_string = format_user_facing_string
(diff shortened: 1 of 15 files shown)


lib/court_data_adaptor/errors.rb
```diff
@@ -31,6 +34,26 @@ def retrieve_user_facing_string
...
+    class ErrorCodeParser
+      def self.call(cda_response)
+        return unless cda_response
+
+        body = cda_response.body
+        error_response = body.is_a?(String) ? JSON.parse(body) : body
+        return unless error_response['error_codes']
+
+        error_response['error_codes'].filter_map { |code| build_message(code) }.join(" ").presence
+      rescue StandardError
```
alexdesi, line 46: is this capturing the `JSON::ParserError`, 
in that case,  maybe we could check `json_string?` , something like this

```
def json_string?(str)
  JSON.parse(str)
  true
rescue JSON::ParserError
  false
end
```

spec/requests/linking/link_defendant_maat_reference_v2_spec.rb
```diff
@@ -37,7 +37,7 @@
     let(:maat_invalid_reference) do
       {
         title: 'Unable to link the defendant using the MAAT ID.',
-        message: 'If this problem persists, please contact the IT Helpdesk on 0800 9175148.'
+        message: 'The MAAT reference you provided is not available to be associated with this defendant.'
```
alexdesi, line 40: We could clarify the message.
Do we mean to say that the maat does not exist?
we could just say: "The MAAT reference you provided does not exist"
or
"The MAAT reference you provided does not exist and therefore cannot be associated with this defendant"

alexdesi (approved):
✅

PR #2979, opened 2025-06-09
ACD-779: New header design

patrick-laa:
#### What
Switch to the MOJ-specific header style.

Also, in debugging this I found some quick wins to improve the dev deployment workflow which I've added to this PR

[Jira ticket](https://dsdmoj.atlassian.net/browse/ACD-779)

Screenshot 2025-06-20 at 09 43 12


yarn.lock
```diff
@@ -2670,11 +2670,16 @@ gopd@^1.0.1, gopd@^1.2.0:
   resolved "https://registry.yarnpkg.com/gopd/-/gopd-1.2.0.tgz#89f56b8217bdbc8802bd299df6d7f1081d7e51a1"
   integrity sha512-ZUKRh6/kUFoAiTAtTYPZJ3hw9wNxx+BIBOijnlG9PnrJsCcSjs1wyyD6vJpaYtgnzDrKYRSqf3OO6Rfa93xsRg==
 
-govuk-frontend@^5.0.0, govuk-frontend@^5.4.0:
+govuk-frontend@^5.0.0:
```
alexdesi, line 2673: should we not use the most recent V.5.4 ? 
... anyway, we can update in a different PR

alexdesi (commented):
it looks good ✅

yarn.lock
patrick-laa: Good spot, MoJ-frontend was requesting "caret 5.0.0" and we were resolving that to  5.4.0. I've updated it so that now when MoJ frontend requests it, we resolve to 5.10.2, which is the latest.

alexdesi (approved):
🚀 ✅

PR #2980, opened 2025-06-10
ACD-807: Verify defendant_link_path working in feature spec

patrick-laa:
#### What
Our CDAPI retirement code went live with a bug in search results where we always linked to the `new_laa_reference` screen rather than the `edit_defendant_screen`, whether or not a defendant was MAAT-linked. While all the relevant code is unit-tested, we didn't have any feature tests that picked up that we didn't have the right decorator in place to make the logic work.


alexdesi (approved):
It looks good! ✅

PR #2985, opened 2025-06-12
Remove "linked" query param and relevant logic.

alexdesi:
#### What
Remove "linked" query param and relevant logic.

#### Ticket
https://dsdmoj.atlassian.net/browse/ACD-820

#### Why
We are going to change the Link/Unlink to sync, so the linked query param will be redundant.


app/controllers/subjects_controller.rb
```diff
@@ -4,16 +4,18 @@ class SubjectsController < ApplicationController
   # do matter, and changing them will affect the breadcrumb order in the UI.
   add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
   add_breadcrumb :application_search_breadcrumb_name, :search_breadcrumb_path
-  before_action :load_and_authorize_application, :load_link
+  before_action :load_and_authorize_application
 
-  def show; end
+  def show
+    @form_model = @link_status.maat_linked? ? load_unlink_attempt : load_link_attempt
```
alexdesi, line 10: This is to make things more explicit and clear to understand, instead of checking the code on the `before_action`.

app/models/court_application_link_status.rb
```diff
@@ -1,25 +1,11 @@
 class CourtApplicationLinkStatus
```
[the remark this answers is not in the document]
alexdesi: yes, I agree, that's not doing much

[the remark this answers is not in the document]
alexdesi: I removed the skinny class,
and added `maat_linked?` in the subjectsummary (+ rspecs)

It should be good to go :)

patrick-laa (approved):
Love it - do we want to hold off on merging until we've made the CDA change to sync?

app/controllers/defendants_controller.rb
```diff
@@ -91,6 +95,10 @@ def unlink_laa_reference_and_redirect
   rescue CourtDataAdaptor::Errors::InternalServerError,
          CourtDataAdaptor::Errors::ClientError => e
     handle_error(e, I18n.t('defendants.unlink.failure'), I18n.t('error.it_helpdesk'))
+  rescue StandardError => e
+    logger.error "MAAT_UNLINK_ERROR: #{e.message}"
```
patrick-laa, line 99: I think it would make more sense to call `handle_error(e, I18n.t('defendants.unlink.unknown_error'), I18n.t('error.it_helpdesk'))` here. That way:
a) we still log the error to Sentry, so we'll get an alert about it, and
b) if by any chance the error has a cda error code in it, we still have a chance of showing a message specific to that error code.

alexdesi: I tried, but  handle_error is only compatible with `CourtDataAdaptor::Errors`, not with StandardError.
Anyway I agree on capturing the exception on Sentry.

patrick-laa (approved):

app/controllers/hearings_controller.rb
```diff
@@ -76,10 +73,6 @@ def log_and_capture_error(exception, log_messsage)
     Sentry.capture_exception(exception)
   end
 
-  def decorate_hearing(undecorated_hearing)
```
alexdesi, line 79: method practically useless

PR #2992, opened 2025-06-18
ACD-828: OAuth tidyup

patrick-laa:
#### What
While investigating some oauth call failures I found some improvements we can make to our OAuth logic. These are:
- Use the singleton pattern to allow us to stop generating brand new access tokens on every page load
- Don't wait until the token has expired before refreshing it - build in a 1-minute buffer
- Don't make each Query class call `refresh_token_if_required` explicitly. Move it into the base class so it's called automatically.
- Stop generating an access token on startup (e.g. when launching `rails c` locally). Instead wait until it's needed, to avoid network issues when a newly minted pod fires up for the first time

#### Ticket

[board ticket description](https://dsdmoj.atlassian.net/browse/ACD-828)


alexdesi (approved):
Nice refactor, I like it! 
✅

PR #2993, opened 2025-06-18
ACD-830: POST instead of GET for searches

patrick-laa:
#### What
POST instead of GET for searches so that we don't put PII into logs.

#### Ticket

[board ticket description](https://dsdmoj.atlassian.net/browse/ACD-830)


app/models/cda/prosecution_case_search.rb
```diff
@@ -0,0 +1,10 @@
+module Cda
+  class ProsecutionCaseSearch < BaseModel
+    self.element_name = "prosecution_case"
+
+    def self.execute(filter_params)
+      http_response = post('/', nil, { filter: filter_params }.to_json)
+      ProsecutionCaseSearchResponse.new(JSON.parse(http_response.body))
+    end
+  end
+end
```
alexdesi, line 10: I am bit confused about this class,
it is a model, but it behaves like a service, 
it uses post but it is actually not creating a new resource 🤔 
... I understand this comment is not helpful, just I am not totally sure about the needs of this refactor,
there wasn't another way to avoid put PII into logs?

app/models/cda/prosecution_case_search.rb
```diff
@@ -0,0 +1,10 @@
+module Cda
+  class ProsecutionCaseSearch < BaseModel
+    self.element_name = "prosecution_case"
+
+    def self.execute(filter_params)
```
alexdesi, line 5: I think we use `.call` instead of execute

patrick-laa: I've tried to fix your comment about "it is a model, but it behaves like a service," bye changing this from `execute` to `create`

alexdesi (approved):
thanks for the clarification! ✅

PR #3000, opened 2025-06-24
ACD-843: Handle multiple responses to URN search

patrick-laa:
#### What
Amazingly, Common Platform can return 2 results for a URN search, because searches are case-insensitive and some URNs differ _only_ in case.


alexdesi (approved):
great 🚀 ✅

PR #3006, opened 2025-06-30
Remove Flash errors and use form validation to show errors.

alexdesi:
#### Ticket
https://dsdmoj.atlassian.net/browse/ACD-852

#### What
- Remove use of Flash error in the laa_references_controller#create
- Refactor and simplify laa_references_controller

### DIFF
--- a/app/controllers/defendants_controller.rb
+++ b/app/controllers/defendants_controller.rb
@@ -87,7 +87,7 @@ class DefendantsController < ApplicationController
     # reason code must be an integer from 1..7
     unlink_attempt_params.merge(username: current_user.username).tap do |attrs|
       attrs[:defendant_id] = defendant.id
-      attrs[:maat_reference] = defendant.maat_reference.to_i
+      attrs[:maat_reference] = defendant.maat_reference
       attrs[:reason_code] = attrs[:reason_code].to_i
       attrs[:reason_code] = nil if attrs[:reason_code].zero?
     end
--- a/spec/features/unlinking_a_defendant_from_maat_spec.rb
+++ b/spec/features/unlinking_a_defendant_from_maat_spec.rb
@@ -45,7 +45,7 @@ RSpec.feature 'Unlinking a defendant from MAAT', :stub_unlink, type: :feature do
   context 'when user views linked defendant' do
     let(:defendant_id) { '41fcb1cd-516e-438e-887a-5987d92ef90f' }
     let(:url) { "defendants/#{defendant_id}/edit?urn=#{case_urn}" }
-    let(:maat_reference) { 2_123_456 }
+    let(:maat_reference) { 2_123_456.to_s }
 
     it 'does not display the MAAT ID field' do
       expect(page).to have_no_field('MAAT ID')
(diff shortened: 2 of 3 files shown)


app/controllers/laa_references_controller.rb
```diff
@@ -22,8 +22,23 @@ def new; end
   def create
     authorize! :create, :link_maat_reference, message: I18n.t('unauthorized.default')
 
-    link_laa_reference_and_redirect && return if @link_attempt.valid?
-    render 'new'
+    @link_attempt.validate!
```
alexdesi, line 25: I recognise that ` @link_attempt` is a pretty bad name for a form 🤔 I'd be happy to rename it.

config/locales/en/cda_errors.yml
```diff
@@ -3,7 +3,9 @@ en:
   cda_errors:
     common_platform_connection_failed: HMCTS Common Platform could not be reached. This may be a temporary error. If this problem persists, please contact the IT Helpdesk on 0800 9175148.
     defendant_id_contract_failure: The defendant you are trying to link is not in a valid state to be linked.
+    internal_server_error: Court Data Adaptor could not be reached. This may be a temporary error. If this problem persists, please contact the IT Helpdesk on 0800 9175148.
     maat_reference_contract_failure: The MAAT reference you provided is not available to be associated with this defendant.
     multiple_maats: The HMCTS record for this defendant cannot be displayed. HMCTS is aware of this issue. The record will be available once this issue has been resolved.
     spaces_in_urn: The HMCTS record for this defendant cannot be displayed. HMCTS is aware of this issue. The record will be available once this issue has been resolved.
     subject_id_contract_failure: The appellant you are trying to link is not in a valid state to be linked.
+    unlink_contract_failure: The request to unlink the Defendant/Appellant was malformed.
```
alexdesi, line 11: the changes in this PR requires error_codes to be added in some API endpoint response. 
`unlink_contract_failure` must be added in the CDA as a possible `error_code`

patrick-laa: Careful, that's not quite how CDA error codes work. If you look in app/models/errors.rb in CDA you'll see that ContractError will generate error codes based on the specific _field_ in the contract that failed. The difficulty is that while it does mention the field in the contract (e.g. `subject_id_contract_failure` or `maat_reference_contract_failure`), it doesn't mention the name of the contract. So you could get a `subject_id_contract_failure` both when linking an appellant and when unlinking an appellant. If you want a different error message from CDA when linking vs unlinking you'll probably want to change the logic to include both field name _and_ contract name, but if you do that you'll need to update the existing keys `defendant_id_contract_failure`, `defendant_id_contract_failure` or `maat_reference_contract_failure` in this file to adopt the new format.

app/controllers/defendants_controller.rb
```diff
@@ -47,6 +56,12 @@ def prosecution_case_reference
 
   private
 
+  def handle_link_failure(message, exception = nil)
+    logger.warn "LINK DEFENDANT FAILURE (params: #{@unlink_attempt.as_json}): #{message}"
```
patrick-laa, line 60: This should say `UNLINK DEFENDANT...` and the method should probably be called `handle_unlink_failure`

app/controllers/defendants_controller.rb
```diff
@@ -47,6 +56,12 @@ def prosecution_case_reference
...
+    @unlink_attempt.errors.add(:maat_reference,
```
patrick-laa, line 61: I'm not sure `maat_reference` is the right field to add this error to here, since there's no form field for it. I'd go either with `:base` or maybe `:reason_code`?

alexdesi: yep, :base sounds good to me!

config/locales/en/cda_errors.yml
alexdesi: That's interesting, for now I'll use existing keys generated by CDA.
We can improve it in others PRs.

patrick-laa (approved):

PR #3008, opened 2025-07-02
ACD-832: Entra SSO

patrick-laa:
SSO via Entra ID.

I added a 'Dev' one-click login because the multi-subdomain dev environment won't really work for Entra ID, and having an easy way to log in as different roles is useful for local development, and necessary for our dockerised E2E tests

Omniauth requires Faraday v2, which is incompatible with the now deprecated FaradayMiddleware, so I removed that.

Before this can be merged we need:
* ~Sign-off on the UI/UX and approach to user management~
* ~Entra credentials for our test, UAT and production environments~ ✅

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -3,11 +3,15 @@ module Users
     skip_authorization_check
 
     def entra
-      user = User.find_by(entra_id: entra_details.uid) || User.find_by(entra_id: nil, email: entra_email)
+      user = User.find_by(entra_id: entra_details.uid)
+      if user # User is found by entra_id, which means they already logged in at least once
+        user.update!(email: entra_email, email_confirmation: entra_email)
+      else
+        user = User.find_by(entra_id: nil, email: entra_email)
+        user&.update!(entra_id: entra_details.uid) # User is found by email and no entra_id, which means it's their first login
+      end
+
       if user
-        user.update!(entra_id: entra_details.uid,
-                     email: entra_email,
-                     email_confirmation: entra_email)
         sign_in user
         redirect_to authenticated_root_path
       else


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    def retrieve_user
+      User.find_by(entra_id: entra_details.uid) ||
+        User.where(entra_id: nil).find_by(email: entra_email)
+    end
+
+    def update_user(user)
+      user.update!(entra_id: entra_details.uid,
+                   email: entra_email,
+                   email_confirmation: entra_email)
+    end
```
alexdesi, line 29: You could inline this or make the method private.

helm_deploy/values/dev.yaml
```diff
@@ -12,7 +12,11 @@ googleAnalyticsId: 'UA-131160087-14'
 displayRawResponses: enabled
 identifier: laa-court-data-ui-app-ingress-laa-court-data-ui-dev-green
 cert: laa-court-data-ui-dev-cert
+# TODO: Before merge, switch back to main dev
```
alexdesi, line 15: remind TODO 😄

spec/features/sign_in_spec.rb
```diff
@@ -55,26 +63,67 @@
...
+    end
+  end
+
+  context 'with an unsuccessful omniauth flow' do
+    before do
+      OmniAuth.config.mock_auth[:entra] = :invalid_credentials
+      click_button 'Sign in with your Ministry of Justice account'
+    end
+
+    it 'invalid username, email or password displayed' do
```
alexdesi, line 111: Active tense instead of passive?
`it 'displays invalid username, email or password'`

alexdesi (approved):
As you wrote, this is not mergeable as it is, but approving with some comments (I'll be off for 2 weeks).

PR #3010, opened 2025-07-03
ACD-863: Cope with maat references that start with A

patrick-laa:
#### What
When users click 'Create link without a MAAT ID', a dummy MAAT ID gets generated that starts with an `A`. When we try to unlink, we need to pass the current MAAT ID to CDA, but `to_i` turns the dummy MAAT ID to a 0. MAAT IDs are strings, not numbers (because they can contain A or Z) so should be treated as such.


alexdesi (approved):
✅ it looks good!

PR #3012, opened 2025-07-07
Paginate users

patrick-laa:
#### What
Add pagination (via govuk_pagination/pagy) to the user list


alexdesi (approved):
It looks good ✅ 
Even though, if we remove all the inactive users maybe we wouldn't  need to implement pagination and search.

PR #3013, opened 2025-07-08
ACD-869: Add a prometheus exporter container to worker pods

patrick-laa:
#### What
Worker pods try to export their metrics to prometheus on localhost:9394 just like web pods, but unlike web pods, there's nothing listening on localhost:9394. This PR adds the thing that listens on that port and exports the data to Prometheus. Not only does this get rid of the error message clogging up the logs, it means we get pod health metrics from worker pods added to the data that makes it into Prometheus.


alexdesi (approved):
good idea! ✅

PR #3017, opened 2025-07-09
ACD-873: Format all user-facing dates appropriately

patrick-laa:
#### What
https://dsdmoj.atlassian.net/browse/ACD-873
Use format "1 January 1900" instead of "01/01/1900"

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,6 +47,7 @@ module LaaCourtDataUi
 
     config.active_support.to_time_preserves_timezone = :zone
 
-    Time::DATE_FORMATS[:date_only] = "%b %d, %Y"
+    Time::DATE_FORMATS[:date_only] = "%-d %B %Y"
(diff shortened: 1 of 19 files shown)


config/application.rb
```diff
@@ -47,6 +47,7 @@ class Application < Rails::Application
 
     config.active_support.to_time_preserves_timezone = :zone
 
-    Time::DATE_FORMATS[:date_only] = "%b %d, %Y"
+    Time::DATE_FORMATS[:date_only] = "%-d %B %Y"
+    Date::DATE_FORMATS[:date_only] = "%-d %B %Y"
```
alexdesi, line 51: Nice refactor!
but `:date_only` does not say much about the date,
we can have date only in many different format.
Should we rename it to something like:
`:day_month_year`
`:readable_date`
`:long_date`
...
up to you :)

patrick-laa:
Closing as we have decided to stick with the current system.

PR #3021, opened 2025-07-14
ACD-880: Sort users by name

patrick-laa:
#### What
On the user list pages, sort users by last name then first name, so the order would be:

Aaa Aaa
Bbb Aaa
Aaa Bbb
Bbb Bbb


alexdesi (approved):
the code looks good ✅ 
I just wonder if the sorting is evident to the users (since the last name appears on the right)

patrick-laa:
I will check with Chuan that he's happy with the sort order

PR #3022, opened 2025-07-14
Only monitor on merge

patrick-laa:
#### What
Scan with Snyk on every PR, but only do ongoing monitoring on merges.


alexdesi (approved):
it looks good ✅ 
ps. check comment in previous PR.

PR #3025, opened 2025-07-16
ACD-886: Separate heading from title when heading has PII

patrick-laa:
#### What
When the page heading has PII, provide a separate page title so that the PII doesn't get sent to Google Analytics.

### DIFF
--- a/app/helpers/govuk_design_system_helper.rb
+++ b/app/helpers/govuk_design_system_helper.rb
@@ -7,14 +7,14 @@
 require 'gds_design_system_breadcrumb_builder'
 
 module GovukDesignSystemHelper
-  def govuk_page_title(heading = nil, caption = nil, options = {})
-    title = options.delete(:title)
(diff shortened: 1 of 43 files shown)


app/helpers/govuk_design_system_helper.rb
```diff
@@ -7,13 +7,14 @@
 require 'gds_design_system_breadcrumb_builder'
 
 module GovukDesignSystemHelper
-  def govuk_page_title(title = nil, caption = nil, tag_options = {})
-    content_for :page_title, page_title(title, caption)
+  def govuk_page_title(heading = nil, caption = nil, options = {})
+    title = options.delete(:title)
```
alexdesi, line 11: it looks a bit strange that `title`, wich is one important param of this method, is "hidden" inside `options`
is not the caption something optional?

alexdesi (approved):
it looks good ✅ 
just a minor comment to clarify title and caption

PR #3027, opened 2025-07-17
ACD-889: Retry flaky JS specs

patrick-laa:
#### What
Our JS specs are increasingly flaky, failing with errors like:
```
unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}
```
and
```
javascript error: Cannot read properties of undefined (reading 'runPartial')
```

This appears to be a general degradation of Selenium. See for example https://github.com/SeleniumHQ/selenium/issues/15401.

We could solve this by eliminating all our JS specs, but at the cost of removing all our automated accessibility testing. All our `expect(page).to be_accessible` statements require a JS-capable test driver

Since we can't currently stop these specs being flaky, nor do we want to remove them, we can mitigate the flakiness by automatically retrying JS specs that fail using `rspec-retry`. This means that if a JS spec fails we will re-run the spec up to twice more, and will mark the spec as a success and move on if any of the retries pass.

Obviously this would cause problems if we had intermittent actual bugs, as our tests wouldn't catch them, but given how little actual JS we use the likelihood of a real bug that only occurs occasionally in JS specs is low.


alexdesi (approved):
good idea ✅

PR #3028, opened 2025-07-17
ACD-890: Sanitise URL params we pass on to CDA

patrick-laa:
#### What
VCD is breaking when the URN has a space in it, but escaping is also important to prevent malicious requests being passed on to CDA

### DIFF
--- a/app/models/cda/defendant.rb
+++ b/app/models/cda/defendant.rb
@@ -4,8 +4,8 @@ module Cda
   class Defendant < BaseModel
     def self.find_from_id_and_urn(defendant_id, urn)
       find(:one,
-           from: "/api/internal/v2/prosecution_cases/#{make_safe_for_path(urn)}/" \
-                 "defendants/#{make_safe_for_path(defendant_id)}")
+           from: "/api/internal/v2/prosecution_cases/#{safe_path(urn)}/" \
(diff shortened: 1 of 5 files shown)


app/models/cda/defendant.rb
```diff
@@ -3,7 +3,9 @@
 module Cda
   class Defendant < BaseModel
     def self.find_from_id_and_urn(defendant_id, urn)
-      find(:one, from: "/api/internal/v2/prosecution_cases/#{urn}/defendants/#{defendant_id}")
+      find(:one,
+           from: "/api/internal/v2/prosecution_cases/#{make_safe_for_path(urn)}/" \
+                 "defendants/#{make_safe_for_path(defendant_id)}")
```
alexdesi, line 8: I was thinking, can we sanitize the entire path, instead of urn and defendant_id separetely?

alexdesi (approved):
it looks good, ony minor comments ✅

app/models/cda/defendant.rb
patrick-laa: Sadly not, because different rules apply to identifiers than the rest of the path. E.g. we absolutely have to make sure there are no `/` characters in the URN, as that would cause us to call a different endpoint to the one intended. But there _can_ be slashes in the rest of the path.

PR #3038, opened 2025-07-23
ACD-298: Get rid of all things cypress

patrick-laa:
#### What
We're not using it, and all it really tested was the cookie banner anyway.


alexdesi (approved):
Nice cleaning! ✅

PR #3042, opened 2025-07-24
ACD-902: Load offences from prosecution case

patrick-laa:
#### What
https://dsdmoj.atlassian.net/browse/ACD-902

Common Platform appeal offences don't have plea information. But since we want to display the exact same offence information as the defendant page, i.e using the prosecution case offence payload data, we can simply reuse that table (and its async load mechanism) to show the prosecution case offence data, which has the full plea history.


app/controllers/subjects_controller.rb
```diff
@@ -88,6 +96,15 @@ def load_link_attempt
...
 
+  def load_defendant
+    prosecution_case = CourtDataAdaptor::CaseSummaryService.call(@application.case_summary.prosecution_case_reference)
+    @defendant = prosecution_case.defendant_summaries.find do |defendant|
+      defendant.application_summaries.any? do |as|
+        as.id == @application.application_id
+      end
+    end
+  end
+
```
alexdesi, line 107: I am not a big fan of this `load_...` pattern:
- It obfuscates the vars that are passed into the views
- It adds additional logic in the controllers 
- It's hard to test

An alternative is to just set the var in the action:
```
@defendant = get_defendant_by_prosecution_case(@application.case_summary.prosecution_case_reference)
```
and move `get_defendant_by_prosecution_case` to a "better" place, maybe inside the `prosecution_case` ? 🤔 
`prosecution.get_defendant`

app/views/defendants/_offences.html.haml
```diff
@@ -12,6 +13,7 @@
       %th.govuk-table__header{ scope: 'col', class: 'govuk-!-width-one-quarter' }= t('search.result.offence.mode_of_trial')
   %tbody.govuk-table__body
     - decorate_each(defendant.offence_summaries, OffenceDecorator) do |offence|
+      - next if offence_ids&.exclude?(offence.id)
```
alexdesi, line 16: It'd be great to exclude this from inside `defendant.offence_summaries`
Maybe using a new method inside the `defendant`?

app/controllers/subjects_controller.rb
patrick-laa: Fair point. I've added `.defendant` as a method on ApplicationSummary, and at that point we don't need to set an ivar at all, we can simply call the method when needed in the view.

PR #3051, opened 2025-08-08
ACD-952: Per-user feature flags for appeals

patrick-laa:
#### What
Make it possible to enable feature flags, and specifically the appeals feature flag, for specific users, with a mechanism in the UI to enable this.


muthusubra:
Looks okay to me. Just one comment. Would it be better to use enum for feature_flag as against an array, to restrict the values to a predefined set and possibly enum might be easier to handle and query?

muthusubra (approved):

PR #3062, opened 2025-08-22
Add applicant counsels and respondent counsels to the attendees panel.

alexdesi:
#### What
The attendees' panel does not show the advocates (counsels) in the hearing view in case the hearing is related to a appeal

### DIFF
--- a/app/views/hearings/_attendees.html.haml
+++ b/app/views/hearings/_attendees.html.haml
@@ -4,11 +4,12 @@
   .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
     = t('hearings.show.attendees.defendants')
   %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "defendants" }
-    - if hearing_details&.prosecution_cases&.any?
-      - cases = hearing_details.prosecution_cases
-      = safe_join(cases.map { |cs| cs.defendants.map { |df| "#{df.defendant_details.person_details.first_name} #{df.defendant_details.person_details.last_name}" } }, tag.br)
+    -# TODO: get the name from the prosecution case summary payload
+    - if prosecution_case.defendant_summaries.any?
+      = safe_join(prosecution_case.defendant_summaries.map{ |d| [d.first_name, d.middle_name, d.last_name].reject(&:blank?).join(' ') }, tag.br )
     - else
       = t('generic.not_available')
+
   .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
     = t('hearings.show.attendees.defence')
   %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "defence" }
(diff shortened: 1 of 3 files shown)


app/views/hearings/_attendees.html.haml
```diff
@@ -1,28 +1,57 @@
 %div{ class: "govuk-!-margin-bottom-9" }
   %h2.govuk-heading-l
     = t('hearings.show.attendees.heading')
+
+  -# TODO: refactor: add method is_an_appeal? in the decorator
+  - is_an_appeal = hearing_details && hearing_details.hearing_type.description.include?('Appeal')
```
patrick-laa, line 6: I don't think you should be searching for specific words in strings here. For one thing, this will immediately break when we start adding breaches etc and court applications that aren't appeals. For another, the text in the description is only indirectly related to the problem you're trying to solve.

The problem you have is that sometimes you'll be on this screen and `hearing_details.prosecution_cases` will be blank, in which case you need to populate this screen from appeal details rather than prosecution case details. So your check should be more like `is_an_appeal = hearing_details.prosecution_cases.blank?`

app/views/hearings/_attendees.html.haml
```diff
@@ -1,28 +1,57 @@
 %div{ class: "govuk-!-margin-bottom-9" }
   %h2.govuk-heading-l
     = t('hearings.show.attendees.heading')
+
+  -# TODO: refactor: add method is_an_appeal? in the decorator
```
patrick-laa, line 5: Yes, we should do this before merging

app/views/hearings/_attendees.html.haml
```diff
@@ -1,28 +1,57 @@
...
-      - cases = hearing_details.prosecution_cases
-      = safe_join(cases.map { |cs| cs.defendants.map { |df| "#{df.defendant_details.person_details.first_name} #{df.defendant_details.person_details.last_name}" } }, tag.br)
-    - else
-      = t('generic.not_available')
-  .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
-    = t('hearings.show.attendees.defence')
-  %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "defence" }
-    - if hearing_details&.defence_counsels&.any?
-      = safe_join([hearing.defence_counsels_list], tag.br)
+    - if is_an_appeal
```
patrick-laa, line 9: It's feeling like maybe rather than having multiple `if is_an_appeal` checks we should have 2 partials, and do a single `if` check and render the appropriate partial based on that?

app/decorators/cda/hearing_decorator.rb
```diff
@@ -50,6 +50,10 @@ def judiciary_list
       safe_join(hearing.judiciary.map { |jd| "#{jd.title} #{jd.first_name} #{jd.last_name}" }, tag.br)
     end
 
+    def appeal?
+      hearing.hearing_type.description.downcase.include?('appeal') if loaded? && hearing.hearing_type.present?
```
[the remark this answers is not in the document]
patrick-laa: I think we can solve this: we don't actually care whether it's an appeal or not. We care about whether the hearing has defendants/prosecutors or applicants/respondents. So if you want to bullet-proof this, you should do:

```
is_a_prosecution_case = hearing_details.prosecution_cases.any?

...

if is_a_prosecution_case
  # Do what we're doing already
else
  # Do the new thing.
end
```

For the time being we can assume that any court application hearing visible in VCD will be an appeal type, but when breaches is switched on we'll need to start working out if it's an appeal or a breach and show "Appellant" or "Applicant" as appropriate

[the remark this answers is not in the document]
alexdesi: Good point, it's better to see the thing in the other way :)
👍

PR #3063, opened 2025-08-22
Add appeals-specific feedback banner for appeals launch

alexdesi:
#### What
Add appeals-specific feedback banner for appeals launch

#### Ticket
https://dsdmoj.atlassian.net/browse/ACD-962

#### Why
To gather feedback from caseworkers.


muthusubra (approved):

app/views/_feedback_on_appeals_to_crown_court.html.haml
```diff
@@ -0,0 +1,10 @@
+.govuk-notification-banner{ role: "region", aria: { labelledby: "govuk-notification-banner-title" }, data: { module: "govuk-notification-banner" } }
+  .govuk-notification-banner__header
+    %h2.govuk-notification-banner__title#govuk-notification-banner-title
+      Important
+  .govuk-notification-banner__content
+    %p.govuk-notification-banner__heading
+      Give
```
patrick-laa, line 7: It would be really nice if we could be consistent about keeping copy in the i18n yml files (I agree that since this will never be translated, we don't really need them, but it's a pain when almost all the copy is there and then some is hiding in the view files

app/views/_feedback_on_appeals_to_crown_court.html.haml
```diff
@@ -0,0 +1,10 @@
+.govuk-notification-banner{ role: "region", aria: { labelledby: "govuk-notification-banner-title" }, data: { module: "govuk-notification-banner" } }
```
patrick-laa, line 1: You could save some hassle if you use `= govuk_notification_banner(...)` here instead of using the HTML by hand - we added the govuk-components gem recently

.tool-versions
```diff
@@ -0,0 +1,2 @@
+nodejs 24.4.0
```
patrick-laa, line 1: What does this do?

alexdesi: It's a way to specify the version of the tools we are using,
it's like `.ruby-version` , but more generic
we use it in CDA,
it is used by `asdf` and other versioning tools

app/views/_feedback_on_appeals_to_crown_court.html.haml
alexdesi: I see your point, but
- It’s temporary.
- Only one language is supported.
- Adding it to en.yml would just add noise you’ll delete later.

app/views/_feedback_on_appeals_to_crown_court.html.haml
alexdesi: Actually, that's not easy to reuse, since in our case, the heading has a link inside it.
The example in:  https://govuk-components.netlify.app/components/notification-banner/#output-rendered-notification-banner-with-heading is similar, but not flexible enough (the link can be placed only at the end)

PR #3064, opened 2025-08-26
ACD-909: (Appeals V2) New FF and new tags in UI

patrick-laa:
#### What
Add new Appeals v2 feature flag
Use it to display alternate tags in UI


alexdesi (approved):
The code is fine,
my only concern is that we are putting behind a feature flag these small UI improvements.
They are messing up our code, and I think,
being UI improvements, they should be part of appeal.
Of course, for bigger changes it's okay to use a feature flag for the release.

PR #3066, opened 2025-08-26
ACD-911: (Appeals V2) Remove appeals from search results page

patrick-laa:
Remove appeals from search results page


alexdesi (approved):
It looks good!
Please check if this will break some of the E2E (Playwright) tests.

PR #3067, opened 2025-08-26
ACD-912: (Appeals v2) Related applications tab

patrick-laa:
If the FF is enabled, show link to related applications tab


app/models/cda/court_application.rb
```diff
@@ -3,6 +3,12 @@ class CourtApplication < BaseModel
     has_one :subject_summary, class_name: 'Cda::SubjectSummary'
     has_many :hearing_summary, class_name: 'Cda::ApplicationHearing'
 
+    def self.from_urn(urn)
```
alexdesi, line 6: since we have `CourtApplication.find`
maybe call this 
`CourtApplication.find_from_urn` for consistency?

spec/features/court_applications_spec.rb
```diff
@@ -26,6 +26,23 @@
     expect(page).to have_content "There was a problem getting the information you requeste"
   end
 
+  context 'when the FF is enabled' do
+    before do
+      allow(FeatureFlag).to receive(:enabled?).with(:maintenance_mode).and_return(false)
```
alexdesi, line 31: I thought this was maintenance_mode was false by default

app/controllers/prosecution_cases_controller.rb
```diff
@@ -5,14 +5,19 @@ class ProsecutionCasesController < ApplicationController
 
   add_breadcrumb :search_filter_breadcrumb_name, :new_search_filter_path
   add_breadcrumb :search_breadcrumb_name, :search_breadcrumb_path
+  before_action :add_final_breadcrumb
```
alexdesi, line 8: nitpick: I think this is not a special case, 
You could inline the breadcrumb here, or maybe put all the breadcrumbs inside the method
`before_action :add_breadcrumbs`

alexdesi (approved):
it's really good!
I've just added minor comments.

Note: the new release of gem `pagy` is going to be a major change, it'll be released soon, the rc3 it's been released. 
We should update it soon, in a different pr.

spec/features/court_applications_spec.rb
patrick-laa: It is, I was trying to deal with RSpec complaining if an `allow`ed method gets called with different arguments, but I realised I can do it better. Now updated.

alexdesi:
great! Thanks for the changes 🚀

PR #3068, opened 2025-08-26
ACD-915: (Appeals v2) Move unlink form out of collapsed 'details' component

patrick-laa:
https://dsdmoj.atlassian.net/browse/ACD-915


app/views/defendants/_unlink_warning.html.haml
```diff
@@ -0,0 +1,7 @@
+- classes ||= ''
```
alexdesi, line 1: this is smart, but not too easy to maintain, 
can we add the `govuk-!-margin-bottom` in the caller of the partial? Maybe with a span or something?

alexdesi (approved):
It looks good,
but I am not convinced about the passing in the `classes` as partial argument

PR #3069, opened 2025-08-26
ACD-961: (Appeals v2) URN on subject page

patrick-laa:
https://dsdmoj.atlassian.net/browse/ACD-961


alexdesi (approved):
It looks great 👍

PR #3071, opened 2025-08-27
Fix attendees panel - Refactored

alexdesi:
#### What
This PR fixes the Attendees panel
This is a second version, with cleaner code than the original PR.

### DIFF
--- a/app/views/hearings/show.html.haml
+++ b/app/views/hearings/show.html.haml
@@ -12,8 +12,7 @@
     - is_prosecution_case = @hearing.try(:hearing)&.prosecution_cases&.any?
     - if is_prosecution_case
       = render partial: 'attendees_on_prosecution_cases', locals: { hearing: @hearing,
-                                                                    hearing_details: @hearing.try(:hearing),
-                                                                    prosecution_case: @prosecution_case }
+                                                                    hearing_details: @hearing.try(:hearing) }
     - else
       = render partial: 'attendees_on_court_applications', locals: { hearing: @hearing,
                                                                      hearing_details: @hearing.try(:hearing),
--- a/app/views/hearings/_attendees_on_court_applications.haml
+++ b/app/views/hearings/_attendees_on_court_applications.haml
@@ -10,23 +10,21 @@
     - else
       = t('generic.not_available')
 
-  - if hearing_details.respond_to?(:applicant_counsels)
-    .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
-      = t('hearings.show.attendees.appellant_advocates')
-    %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "appellant" }
-      - if hearing_details.applicant_counsels.any?
-        = hearing.applicant_counsels_list
-      - else
-        = t('generic.not_available')
+  .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
+    = t('hearings.show.attendees.appellant_advocates')
+  %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "appellant" }
+    - if hearing_details.try(:applicant_counsels)&.any?
+      = hearing.applicant_counsels_list
+    - else
+      = t('generic.not_available')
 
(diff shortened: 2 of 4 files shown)


app/views/hearings/show.html.haml
```diff
@@ -9,7 +9,15 @@
...
                                            hearing_day: @hearing_day }
 
-    = render partial: 'attendees', locals: { hearing: @hearing, hearing_details: @hearing.try(:hearing) }
+    - is_prosecution_case = @hearing.try(:hearing)&.prosecution_cases&.any?
+    - if is_prosecution_case
+      = render partial: 'attendees_on_prosecution_cases', locals: { hearing: @hearing,
+                                                                    hearing_details: @hearing.try(:hearing),
+                                                                    prosecution_case: @prosecution_case }
+    - else
+      = render partial: 'attendees', locals: { hearing: @hearing,
```
patrick-laa, line 18: I think there's a strong case for renaming this file to "_attendees_on_court_applications.html.haml" rather than just "attendees"

app/views/hearings/_attendees_on_prosecution_cases.html.haml
```diff
@@ -0,0 +1,35 @@
+%div{ class: "govuk-!-margin-bottom-9" }
+  %h2.govuk-heading-l
+    = t('hearings.show.attendees.heading')
+
+  .govuk-heading-s{ class: "govuk-!-margin-bottom-1" }
+    = t('hearings.show.attendees.defendants')
+  %p.govuk-body{ class: "govuk-!-margin-bottom-2", id: "defendants" }
+    - if prosecution_case.defendant_summaries.any?
+      = safe_join(prosecution_case.defendant_summaries.map(&:name), tag.br)
```
patrick-laa, line 9: Careful, you've introduced a bug here: Current behaviour is that this panel only shows defendants _who attended the hearing_. This will replace that with _all defendants_. We don't want that.

spec/views/hearings/_attendees_on_prosecution_cases.html.haml_spec.rb
```diff
@@ -1,14 +1,19 @@
 # frozen_string_literal: true
 
-RSpec.describe 'hearings/_attendees', type: :view do
+RSpec.describe 'hearings/_attendees_on_prosecution_cases', type: :view do
```
patrick-laa, line 3: It'd be great to get at least _some_ test coverage of both partials.

app/views/hearings/_attendees_on_prosecution_cases.html.haml
alexdesi: nice catch!
so, since this case is a prosecution_cases (not an appeal), I think we could revert this to the: 
`if hearing_details&.prosecution_cases&.any?
      - cases = hearing_details.prosecution_cases
      = safe_join(... )
      
 `
 for the view `_attendees` (to be renamed to `_attendees_on_court_applications` )
it shows all the defendants for the prosecution_case:

```
     - if prosecution_case.defendant_summaries.any?
      = safe_join(prosecution_case.defendant_summaries.map(&:name), tag.br)
 ```
 this should be fine,
 since there are not hearing involved, we take all the defendants in the defendant_summaries 
  does it sound correct?

patrick-laa: Yes, that sounds right

spec/views/hearings/_attendees_on_prosecution_cases.html.haml_spec.rb
alexdesi: I agree, but I might not have time for now.
That's the first thing I do after  deploy (if we just want to deploy now),
In any case, I am manually testing this scenario.

alexdesi: Update: I wrote an RSpec for the attendees_on_court_application too.

patrick-laa (approved):

PR #3077, opened 2025-09-02
ACD-934/5: Breaches/POCA

patrick-laa:
#### What
Add a new 'Location' column
Make 'Appellant' into 'Appellant/Respondent'

### DIFF
--- a/app/views/_feedback_on_appeals_to_crown_court.html.haml
+++ b/app/views/_feedback_on_appeals_to_crown_court.html.haml
@@ -1,4 +1,5 @@
-- if application.appeal?
+- application ||= nil
+- if !application || application.appeal?
   .govuk-notification-banner{ role: "region", aria: { labelledby: "govuk-notification-banner-title" }, data: { module: "govuk-notification-banner" } }
     .govuk-notification-banner__header
       %h2.govuk-notification-banner__title#govuk-notification-banner-title


app/models/cda/application_hearing.rb
```diff
@@ -12,5 +12,9 @@ def hearing_days
     def defence_counsels_on(date)
       defence_counsels.select { it.attended_on?(date) }
     end
+
+    def location
```
alexdesi, line 16: I am not entirely convinced that naming this method `location` is the best choice, it's in the model and represents our domain/business logic, not presentation concerns.
Also, it's so generic,  if we search for "location" we'll get hundreds of occurrences.
An alternative might be `court_centre_name`, `court_centre_location`, ...

app/models/cda/court_application.rb
```diff
@@ -23,6 +23,10 @@ def result_string
       Cda::CourtApplicationResultStringService.call(self)
     end
 
+    def location
+      hearing_summary.map(&:location).uniq.to_sentence
```
alexdesi, line 27: silly question time: 
On a quick search for `location`, it's not clear where`location` does come from 🤔 
.. I guess it's the new method on the `application_hearing`.
Perhaps a unit test could clarify things.

alexdesi (approved):
Nice one, 
minor comments to make simpler to work with `location`

app/models/cda/application_hearing.rb
```diff
@@ -12,5 +12,13 @@ def hearing_days
     def defence_counsels_on(date)
       defence_counsels.select { it.attended_on?(date) }
     end
+
+    def jurisdiction
```
alexdesi, line 16: 👍

PR #3078, opened 2025-09-02
ACD-935: (Breaches/POCA) Add breach-specific UI details

patrick-laa:
Use the application_category, that will be either "appeal", "breach" or "poca" to show category-specific details


app/models/cda/court_application.rb
```diff
@@ -29,6 +31,22 @@ def defendant
...
+    def appeal?
+      application_category == "appeal"
+    end
+
+    def breach?
+      application_category == "breach"
+    end
+
+    def maat_reference
+      linked_maat_id unless linked_maat_id&.starts_with?(DUMMY_MAAT_PREFIX)
```
alexdesi, line 43: That's totally fine,
If we want to make it more junior-friendly we could add a guard clause with if and explicit return:
```
  return nil if linked_maat_id&.starts_with?(DUMMY_MAAT_PREFIX)
  linked_maat_id
```
Still, this makes our life hard: we don't know if `linked_maat_id` is nil or if it is a dummy. In the past, cases like this were not easy to debug. (but I guess there is not much we can do for now)

app/services/cda/court_application_result_string_service.rb
```diff
@@ -1,15 +1,18 @@
...
-    def initialize(application_summary)
+    def initialize(application_summary, proceedings_must_be_concluded:)
       @application_summary = application_summary
+      @proceedings_must_be_concluded = proceedings_must_be_concluded
     end
 
     def call
-      return not_available unless application_summary.subject_summary.proceedings_concluded
+      if proceedings_must_be_concluded && !application_summary.subject_summary.proceedings_concluded
+        return not_available
```
alexdesi, line 14: That's very readable :)

alexdesi (commented):
it gooks good,
if I am not wrong, what's missing is probably a couple of tests if application is appeal or breach?
(not sure about the kind of test, I just notice that there are not any)

app/models/cda/court_application.rb
```diff
@@ -3,6 +3,8 @@ class CourtApplication < BaseModel
     has_one :subject_summary, class_name: 'Cda::SubjectSummary'
     has_many :hearing_summary, class_name: 'Cda::ApplicationHearing'
 
+    DUMMY_MAAT_PREFIX = "Z".freeze
```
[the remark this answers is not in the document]
patrick-laa: Ah, that's a slightly different thing. In Ruby 3.4, without the `freeze` call, the best it will do is emit a warning IF `Warning[:deprecated] = true` is set. Whereas with the freeze call it will actually raise a `FrozenError` error

patrick-laa:
@alexdesi OK, I've added some feature specs to the breach details, subject details and hearing details pages to check that we've got the correct content on the page.

PR #3095, opened 2025-09-25
ACD-995: Admin-viewable page of linking stats

patrick-laa:
#### What
A page that only admins can access that lets users search specific date ranges and see statistics pertaining to linking and unlinking for that range, as well as previous comparable ranges

### DIFF
--- a/app/views/stats/show.html.haml
+++ b/app/views/stats/show.html.haml
@@ -56,6 +56,8 @@
   .govuk-grid-column-two-thirds
     %h2.govuk-heading-l{ class: 'govuk-!-margin-top-5 ' }
       = t('stats.show.previous_periods')
+    %p.govuk-hint
+      = t('stats.show.previous_ranges.hint')
     %table.govuk-table{ class: "govuk-!-margin-bottom-4" }
       %caption.govuk-table__caption.govuk-visually-hidden
         = t('stats.show.previous_ranges.caption')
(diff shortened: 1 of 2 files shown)


app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,25 @@
+class StatsController < ApplicationController
+  before_action :authorize_access
+
+  def new
+    @model = StatRange.new
+  end
+
+  def create
```
alexdesi, line 8: This action does not behave like `create`,
Should we use an `index`? 🤔   (which is used to list and search resources, as happens in this case)

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    from = DateFieldCollection.new(params.require(:stat_range), "from")
+    to = DateFieldCollection.new(params.require(:stat_range), "to")
+    @model = StatRange.new(from:, to:)
+
+    return render :new unless @model.valid?
+
+    @collection = Cda::LinkingStatCollection.find_from_range(@model.from.to_s, @model.to.to_s)
+
+    render :show
```
alexdesi, line 17: nit:
Instead of a clause guard, just a boring Rails pattern here?
```
if @model.valid?
   @collection = Cda::LinkingStatCollection.find_from_range(@model.from.to_s, @model.to.to_s)
    render :show
else
  render :new
end
```

patrick-laa:
@alexdesi this is ready for a re-review when you're ready

PR #3206, opened 2026-04-13
ACD-1361: Add link migrated cases page with feature flag

alexdesi:
## Summary

Adds a new "Link migrated cases" page accessible only to caseworkers, gated behind a feature flag so it can be enabled per environment without a code deployment.

- New `LinkMigratedCasesController` with caseworker-only access enforced via CanCanCan
- Navigation link added to the lower nav (conditionally rendered when feature flag is active)
- Feature flag `link_migrated_cases` configured in Helm values for dev, test, UAT, and production environments
- Request, ability, and navigation specs added


sophiebrown01 (approved):
All looks good to me 🎉

PR #3208, opened 2026-04-15
ACD-1380: Handle nil defendant date of birth

alexdesi:
## Summary

When viewing a court application, the page crashed if the defendant's date of birth was missing from the CDA response.

### DIFF
--- a/app/models/cda/subject_summary.rb
+++ b/app/models/cda/subject_summary.rb
@@ -2,6 +2,11 @@ module Cda
   class SubjectSummary < BaseModel
     has_many :offence_summaries, class_name: 'Cda::OffenceSummary'
 
+    # Attributes from CDA payload:
+    #   subject_id, date_of_next_hearing, defendant_asn, defendant_dob,
+    #   defendant_first_name, defendant_last_name, master_defendant_id,
+    #   offence_summary, proceedings_concluded, organisation_name, representation_order.
+
     def name
       [defendant_first_name, defendant_last_name].join(" ")
     end
--- a/app/models/cda/subject_summary.rb
+++ b/app/models/cda/subject_summary.rb
@@ -7,7 +7,7 @@ module Cda
     end
 
     def date_of_birth
-      defendant_dob.to_date
+      defendant_dob.presence&.to_date
     end
   end
 end
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,6 +24,7 @@ default: &default
 development:
   <<: *default
   database: laa_court_data_ui_development
+  host: localhost
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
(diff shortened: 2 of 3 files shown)


app/models/cda/subject_summary.rb
```diff
@@ -7,7 +7,7 @@ def name
     end
 
     def date_of_birth
-      defendant_dob.to_date
+      defendant_dob.presence&.to_date
```
alexdesi, line 15: This is the bug fix!

config/database.yml
```diff
@@ -24,6 +24,7 @@ default: &default
 development:
   <<: *default
   database: laa_court_data_ui_development
+  host: localhost
```
alexdesi, line 27: These changes are unrelated, but good to have.

spec/features/court_applications_spec.rb
```diff
@@ -61,6 +61,14 @@
     end
   end
 
+  scenario 'I view an application where the defendant date of birth is missing' do
+    allow_any_instance_of(Cda::SubjectSummary).to receive(:date_of_birth).and_return(nil)
```
alexdesi: this simulates the bug: date_of_birth is nil

spec/features/court_applications_spec.rb
```diff
@@ -61,6 +61,14 @@
...
+    sign_in user
+    visit court_application_path(found_court_application_id)
+    expect(page).to have_content "Mauricio Rath"
+    expect(page).to have_no_content "06/05/1994"
```
sophiebrown01: I know we have the date of birth listed in the other scenario _I view the application details_ but I'm also wondering if we should have a specific test where we explicitly set `allow_any_instance_of(Cda::SubjectSummary).to receive(:date_of_birth).and_return("06/05/1994")`.

I say this because I don't think that expecting the page to have no content "06/05/1994" isn't actually testing that the date of birth isn't populated, it's just testing that we don't have that content. If we were to be looking inside of the container that should hold the date of birth and check that that is empty, then I don't think we'd need an additional test, but I think the additional test here is the most simple way forwards.

alexdesi: Good spot Sophie! 
yeah I rushed on this,
I also noticed I mocked the method, not the attribute `defendant_dob` 
I'm fixing this

spec/features/court_applications_spec.rb
```diff
@@ -61,6 +61,15 @@
     end
   end
 
+  scenario 'I view an application where the defendant date of birth is missing' do
+    without_partial_double_verification do
+      allow_any_instance_of(Cda::SubjectSummary).to receive(:defendant_dob).and_return(nil)
+    end
+    sign_in user
+    visit court_application_path(found_court_application_id)
+    expect(page.status_code).to eq(200)
```
alexdesi, line 70: This is the only expectation needed. 
If we remove the fix  on the `defendant_dob` , the `visit court_application_path` would blow up (500 error)

spec/features/court_applications_spec.rb
alexdesi: fixed!

sophiebrown01: Thanks Alessandro!

sophiebrown01 (approved):
Looks good! 🚀

PR #3212, opened 2026-04-23
ACD-1360: Add link migrated cases Page

alexdesi:
## Summary

Caseworkers can now view a list of migrated cases from the Link Migrated Cases page.
The table is sortable by case URN, defendant name, Xhibit case number, court name, and mode of trial.

### Code changes

- Introduced `Cda::LinkMigratedCasesCollection` model and `Cda::LinkMigratedCasesService` to fetch and wrap the migrated cases collection from the CDA API
- Changed the `link_migrated_cases` route action from `show` to `index` and updated the controller accordingly
- Added the `index` view with a sortable GOV.UK table; sort direction resets to `asc` when switching columns and toggles when clicking the active column


app/controllers/link_migrated_cases_controller.rb
```diff
@@ -4,12 +4,47 @@ class LinkMigratedCasesController < ApplicationController
...
+                                                   page: page_param, per_page: CASES_PER_PAGE)
+      @cases = @result['results'] || []
+      @pagy = Pagy.new(count: @result['total_results'].to_i, page: page_param, limit: CASES_PER_PAGE)
+      @pending_count = @pagy.count
+    else
+      pending_result = Cda::LinkMigratedCasesService.call(status: TABS['need_linking'], per_page: 1)
+      @pending_count = pending_result['total_results'].to_i
+      @cases = []
+      @pagy = Pagy.new(count: 0, page: 1, limit: CASES_PER_PAGE)
+    end
```
sophiebrown01: Is this block for the pending (not yet tried to be auto-linked) cases? I think the `status` here probably shouldn't be `TABS['need_linking']` but probably `TABS['pending']`? We'll need to revisit this too once we have all the tabs in place to maybe make it a bit more generic.

alexdesi: ~I actually added the other tabs, but they default to 0, there are not calls to filter them for now.~

~My understanding is that 'Need linking' is the UI for the 'pending' (status stored in the DB)~

~In both if/else TABS[tab] and TABS['need_linking'] solve to 'pending' but in the `else` the call exists solely to populate @pending_count for the tab badge, it doesn't care which tab is active, it always needs the pending count. (for the first tab) all the others are set to 0 for now:~

alexdesi: Pagy is the gem that does the pagination, but they are not visible in UAT, we must import more cases,
I'll probably do later, so we can demo it.

alexdesi: I see.
I've added a new tab "Pending".

image

sophiebrown01 (approved):
Amazing work! 🎉

PR #3214, opened 2026-04-30
ACD-1196: Replace accordion with summary cards + New copy buttons

alexdesi:
## Summary

Replaces the GOV.UK accordion component in the hearing court applications partial with summary cards, improving accessibility and consistency with the design system.

The accordion was flagged in the accessibility statement as difficult to use with screen readers (fails WCAG 2.2 success criterion 1.3.2). Summary cards display all content inline without requiring user interaction to expand sections.

### Changes

- `app/views/hearings/_court_applications.html.haml`: replaced `govuk-accordion` with `govuk-summary-card` components. Each card shows the application type as the title, received date top-right in the card header, and the full summary list always visible in the card body. Key and value columns are set to 50/50 width.
- `spec/views/hearings/_court_applications.html.haml_spec.rb`: updated selectors to match the new summary card markup.


app/assets/javascript/copy_button.js
```diff
@@ -0,0 +1,51 @@
+function copyText (textElementId, copyElementId, screenReaderAlertText, originalCopyText = 'Copy') {
```
alexdesi, line 4: This is the function that copies the value into the clipboard,
I DID not write this,
it is part of the MOJ design system: https://design-patterns.service.justice.gov.uk/components/copy-button/#code-tab

sophiebrown01: No problem. Maybe it's worth adding a comment at the top of this file to indicate the source?

alexdesi: yes, that's sensible!
done.

sophiebrown01 (approved):
Looks good 👍 Approving on the basis that the rest of the changes here have already been reviewed by UCD colleauges

PR #3240, opened 2026-06-12
ACD-1420: new linking journey for all the types of Application cases

alexdesi:
Adds the first page of the new linking journey for Breach court applications.


app/controllers/laa_references_controller.rb
alexdesi, line 1: We removed laa_references_controller.rb because its only job (linking a defendant to MAAT) now lives on DefendantsController alongside unlinking. 
The new UI makes link/unlink dedicated pages on the defendant resource (/defendants/:id/link and /unlink), so the separate controller and its /laa_references route were redundant. 

It is much better now:
- One controller owns the journey. Link and unlink are two halves of the same task
- before, link lived in LaaReferencesController and unlink in DefendantsController#edit/update. Now both are actions on the defendant, so related code sits together.
  - Routes match the mental model. /defendants/:id/link and /defendants/:id/unlink read as "act on this defendant," instead of a separate /laa_references resource that
  didn't map to anything users think about.

app/views/subjects/show_link.html.haml
```diff
@@ -30,12 +34,14 @@
           %th.govuk-table__header{ scope: "row" }
             = t(".arrest_summons_number")
           %td.govuk-table__cell
-            = @subject.defendant_asn
-        %tr.govuk-table__row
-          %th.govuk-table__header{ scope: "row" }
-            = t(".plea")
-          %td.govuk-table__cell
-            = @application.result_string
+            -# In case of an appeal, the ASN comes nil from the @subject, so we get it from the @application.defendant
```
sophiebrown01: In the case of a breach, can we also safely get the ASN from the `@subject`?

app/controllers/defendants_controller.rb
```diff
@@ -4,120 +4,136 @@
...
-                        UnlinkAttempt.new
-                      end
+  def load_unlink_attempt
+    reason_code = params.dig(:unlink_attempt, :reason_code).to_i
+    reason_code = nil if reason_code.zero?
+
+    UnlinkAttempt.new(
+      defendant_id: defendant.id,
+      username: current_user.username,
+      reason_code:,
```
sophiebrown01: I think we're missing setting `reason_code` here from lines 112-113?

sophiebrown01 (commented):
Amazing work Alessandro 🥇 Just a few minor comments but otherwise all looks good!

app/controllers/defendants_controller.rb
alexdesi: ha that's ruby syntax, for example:
```
name = "Alice"
age = 28

# Verbose syntax
user_verbose = { name: name, age: age }

# Shorthand syntax
user_shorthand = { name:, age: } 
# Expands to { name: "Alice", age: 28 }
```

Javascript has a similar thing,
anyway, we do not thing two lines to calculate the reason_code, simplified!

app/views/subjects/show_link.html.haml
alexdesi: Yes, breach already uses @subject.defendant_asn.
But I know that this line sounds strange. That branch exists because of a Common Platform bug: it returns defendant_asn: null in the subject_summary for appeals (bug raised by chuan few month ago I think), so we fall back to the defendant record instead. Breach isn't affected by that bug, so no change needed.

app/helpers/defendant_helper.rb
```diff
@@ -2,10 +2,10 @@
...
   def defendant_link_path(defendant, prosecution_case_reference = nil)
-    if defendant.linked?
-      edit_defendant_path(id: defendant.id, urn: prosecution_case_reference)
-    else
-      new_laa_reference_path(id: defendant.id, urn: prosecution_case_reference)
-    end
+    defendant_path(id: defendant.id, urn: prosecution_case_reference)
+  end
+
+  def linking_enabled?
```
alexdesi, line 8: @sophiebrown01 I think we can safely remove the 
linking FeatureFlag 
It's been enabled for all the env for months, and has never been toggled since.

db/seeds/unlink_reasons.csv
```diff
@@ -1,8 +1,9 @@
...
-7,Other,true
+code,description
+1,Linked to right person but wrong case URN
+2,Linked to wrong person and wrong case URN
+3,"Linked to wrong court proceeding (appeal, breach, POCA or case)"
+4,Initially processed on Libra
+5,Duplicate MAAT record for same case
+6,Transfer to another area
+8,There was a system issue
+7,Other
```
alexdesi, line 9: This is actually part of a different PR (which you are reviewing in a separate PR)
I included it here (rebased ) so I could deploy/test together in the test env/uat env.

app/controllers/defendants_controller.rb
```diff
@@ -40,11 +40,9 @@ def link
     validate_link_attempt!
 
     Cda::ProsecutionCaseLaaReference.create!(@form_model.to_link_attributes)
-    @defendant = Cda::Defendant.find_from_id_and_urn(@defendant.id, prosecution_case_reference)
 
     redirect_to defendant_path(@defendant.id, urn: prosecution_case_reference),
-                flash: { success_moj_banner: I18n.t('laa_reference.link.success',
-                                                    maat_id: @defendant.maat_reference) }
+                flash: { success_moj_banner: I18n.t('laa_reference.link.success') }
```
sophiebrown01, line 45: Did we determine that having the MAAT ID in the link/unlink success banners was causing more problems than having it in there / was there any existing user research on this? Happy to go with this, just curious as to the rationale behind the change.

app/controllers/defendants_controller.rb
sophiebrown01: Aha, I didn't spot this. Thanks for clarifying!

app/helpers/defendant_helper.rb
sophiebrown01: Shall we do this in a separate PR? It will make things a bit clearer and cleaner

app/controllers/defendants_controller.rb
alexdesi: yes, specifying the MAAT ID did not add much value to the alert messages, 
and no values at all with the dummy ones.

Laurence confirmed these messages, which do not contain maat ids:

> Case linked successfully.
> Application linked successfully.
> Link removed successfully.
>

alexdesi: Source: https://mojdt.slack.com/archives/C076MS401D5/p1783430073333309?thread_ts=1783096880.462409&cid=C076MS401D5

sophiebrown01 (approved):
All reviewed, all looks good! Great work Alessandro 🥇

app/helpers/defendant_helper.rb
alexdesi: Yes, I agree

PR #3253, opened 2026-06-25
Increase timeout period for liveness and readiness checks

sophiebrown01:
This PR increases the timeout period for liveness and readiness checks for VCD pods. Currently the timeout is not set in code and so defaults to a period of `1s`. During a recent incident, it was observed that VCD pods all restarted simultaneously and the theory for this is that they are tied up on slow request paths to CDA, resulting in requests to the VCD status endpoint taking far longer than `1s`. This commit therefore increases the timeout to a more sensible `10s` limit.

### DIFF
--- a/helm_deploy/templates/deployment.yaml
+++ b/helm_deploy/templates/deployment.yaml
@@ -36,6 +36,7 @@ spec:
                 - name: X-Forwarded-Ssl
                   value: "on"
             initialDelaySeconds: 15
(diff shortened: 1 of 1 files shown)


alexdesi (approved):
✅  It makes sense to me!
The only concern is that 10 sec sounds a bit too much, if it does not respond under 5 sec there must be something wrong (maybe we should give more resources to the pods)

PR #3254, opened 2026-06-26
Fix redirection users/sign_in -> root /

alexdesi:
#### What
Before: users/sign_in treated the string "sign_in" as a ID.
Now: it redirects to /, where the button "sign in" is present.


sophiebrown01 (approved):
Thanks for fixing this Alessandro! 🥇

PR #3256, opened 2026-06-30
Update unlinking reason seeds, and update seeds logic to update the descriptions

alexdesi:
Update unlinking reason seeds, and update seeds logic to update the descriptions

### DIFF
--- a/app/models/unlink_reason.rb
+++ b/app/models/unlink_reason.rb
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
 
 class UnlinkReason < ApplicationRecord
+  OTHER_REASON_CODE = 7
+
   def text_required?
-    code == 7
+    code == OTHER_REASON_CODE
   end
 
   def self.ordered
-    all.sort_by { |r| [r.description == 'Other' ? 1 : 0, r.code] }
+    all.sort_by { |r| [r.text_required? ? 1 : 0, r.code] }
   end
 
   validates :code, presence: true, uniqueness: true
--- a/app/models/unlink_reason.rb
+++ b/app/models/unlink_reason.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class UnlinkReason < ApplicationRecord
+  def self.ordered
+    all.sort_by { |r| [r.description == 'Other' ? 1 : 0, r.code] }
+  end
+
   validates :code, presence: true, uniqueness: true
   validates :description, presence: true, uniqueness: { case_sensitive: false }
 end
--- a/app/models/unlink_reason.rb
+++ b/app/models/unlink_reason.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class UnlinkReason < ApplicationRecord
+  def text_required?
+    code == 7
+  end
+
   def self.ordered
     all.sort_by { |r| [r.description == 'Other' ? 1 : 0, r.code] }
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,42 +10,41 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_08_08_123153) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_01_095152) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
 
   create_table "unlink_reasons", force: :cascade do |t|
     t.integer "code", null: false
-    t.string "description", null: false
-    t.boolean "text_required", default: false, null: false
     t.datetime "created_at", null: false
+    t.string "description", null: false
     t.datetime "updated_at", null: false
     t.index ["code"], name: "index_unlink_reasons_on_code", unique: true
     t.index ["description"], name: "index_unlink_reasons_on_description", unique: true
   end
 
   create_table "users", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.datetime "current_sign_in_at", precision: nil
+    t.inet "current_sign_in_ip"
     t.string "email", default: "", null: false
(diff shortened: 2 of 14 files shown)


db/seeds.rb
```diff
@@ -1,15 +1,5 @@
...
 
-# Seed required data
-#
-
-SEED_FILES = %w[
-  users
-  unlink_reasons
-]
-
-SEED_FILES.each do |file|
```
alexdesi, line 11: Simplification: why use a loop to just load two seeds? 🤷

app/models/unlink_reason.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class UnlinkReason < ApplicationRecord
+  def self.ordered
+    all.sort_by { |r| [r.description == 'Other' ? 1 : 0, r.code] }
+  end
```
alexdesi, line 12: This is a way to customise the sorting, 
the "other reason" must always be at the end (even if there are codes numerically bigger than 7)

app/models/unlink_reason.rb
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 class UnlinkReason < ApplicationRecord
+  OTHER_REASON_CODE = 7
```
alexdesi, line 4: 7 was used around the code base as "magic number"
I gave it the name "OTHER_REASON_CODE", for clarity.

db/seeds/unlink_reasons.csv
```diff
@@ -1,8 +1,9 @@
...
-7,Other,true
+code,description
+1,Linked to right person but wrong case URN
+2,Linked to wrong person and wrong case URN
+3,"Linked to wrong court proceeding (appeal, breach, POCA or case)"
+4,Initially processed on Libra
+5,Duplicate MAAT record for same case
+6,Transfer to another area
+8,There was a system issue
+7,Other
```
alexdesi, line 9: These are the required changes, present in the ticket https://dsdmoj.atlassian.net/browse/LAACONNECT-1446 

Also, I removed the false/true, it is redundant: the only field that requires a reason is "Other" (code 7)

db/schema.rb
```diff
@@ -10,42 +10,41 @@
...
-    t.string "description", null: false
-    t.boolean "text_required", default: false, null: false
     t.datetime "created_at", null: false
+    t.string "description", null: false
     t.datetime "updated_at", null: false
     t.index ["code"], name: "index_unlink_reasons_on_code", unique: true
     t.index ["description"], name: "index_unlink_reasons_on_description", unique: true
   end
 
   create_table "users", force: :cascade do |t|
```
alexdesi, line 26: There are no changes here, 
Just the schema regeneration alphanumerically reordered the attributes

db/schema.rb
```diff
@@ -10,42 +10,41 @@
...
 
-ActiveRecord::Schema[8.0].define(version: 2025_08_08_123153) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_01_095152) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
 
   create_table "unlink_reasons", force: :cascade do |t|
     t.integer "code", null: false
-    t.string "description", null: false
-    t.boolean "text_required", default: false, null: false
```
alexdesi, line 20: `text_required` is redundant!

db/seeds/unlink_reasons.csv
```diff
@@ -1,8 +1,9 @@
...
-1,Linked to wrong case ID (correct defendant),false
-2,Linked to wrong Case ID (incorrect defendant),false
-3,Appeal to Crown Court case,false
+1,Linked to right person but wrong case URN,false
+2,Linked to wrong person and wrong case URN,false
+3,"Linked to wrong court proceeding (appeal, breach, POCA or case)",false
 4,Initially processed on Libra,false
 5,Duplicate MAAT record for same case,false
 6,Transfer to another area,false
+8,There was a system issue,false
```
sophiebrown01, line 8: If we end up getting a lot of these over time we might want to expand this to take in a reason too, optional, to gather more info.

app/models/unlink_reason.rb
sophiebrown01: Is `7` because this is the position of the option in the list of options?

app/models/unlink_reason.rb
```diff
@@ -1,12 +1,14 @@
...
+
   def text_required?
-    code == 7
+    code == OTHER_REASON_CODE
```
sophiebrown01, line 7: I do like this, although if we could remove `7` entirely that'd be even better (because I can foresee new devs in the future asking _why 7?_) 🙂

alexdesi: That is a historical reason, when Patrick and I added "Other" by chance, it was the 7th element.
It must have an ID anyway, but it would have been nice if we had used a bigger number, like **100**, so it stays naturally at the end (instead of using a complex sorting mechanism), and keep it as the last item in the reason codes

We could do that, but it would mess up the stats 🤔
... so not sure at the moment, happy to discuss this further if you have something in mind

db/seeds/unlink_reasons.csv
alexdesi: I think this was overengineered, we've never had this need.
but I see your point 🤔 
Well, if we need it we can revert the code (or rewrite it anyway, it's not complicated)

Also, thinking that from a UX point of view, having an optional textbox appear if you select a radio box, which is not last in the list, may look bad (the other radio buttons should slide down )

app/models/unlink_reason.rb
sophiebrown01: Let's keep as-is for now then but I will add this to the future refactoring list 🙂

sophiebrown01 (approved):
Looks good and thanks for the clarifications 🙂

PR #3262, opened 2026-07-14
Define pod resource requests and limits

sophiebrown01:
This PR defines the pod resource requests and limits for both CPU and memory. Previously these were not defined and would default to the following values:

CPU: request `10m`, limit `1`
Memory: request `100Mi`, limit `1000Mi`

VCD pods have recently been seeing CPU requests exceeding the request limit of `10m` as well as the memory request exceeding `100Mi`, causing the pods to restart. This is likely due to overall load on CDA and needs further investigation, but in the meantime this commit ups the CPU request limit in an attempt to stabilise pods during increased workload.

[Link to task](https://dsdmoj.atlassian.net/browse/LAACONNECT-1460)

### DIFF
--- a/helm_deploy/values/dev.yaml
+++ b/helm_deploy/values/dev.yaml
@@ -5,6 +5,7 @@ replicas:
 image:
   tag: ""
   repository: "nginx"
+resources: {}
 env: dev
 host: dev.view-court-data.service.justice.gov.uk
 cdaHost: https://laa-court-data-adaptor-dev.apps.live.cloud-platform.service.justice.gov.uk/api/internal/v1
--- a/helm_deploy/values/production.yaml
+++ b/helm_deploy/values/production.yaml
@@ -5,6 +5,13 @@ replicas:
 image:
   tag: ""
   repository: "nginx"
+resources:
+  limits:
+    cpu: 1
+    memory: 1000Mi
+  requests:
+    cpu: 100m
+    memory: 500Mi
 env: production
 host: view-court-data.service.justice.gov.uk
 cdaHost: https://court-data-adaptor.service.justice.gov.uk/api/internal/v1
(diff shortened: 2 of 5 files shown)


helm_deploy/values/dev.yaml
```diff
@@ -5,6 +5,7 @@ replicas:
 image:
   tag: ""
   repository: "nginx"
+resources: {}
```
alexdesi, line 8: Should we specify the resources, even if in dev? 🤔

helm_deploy/values/production.yaml
```diff
@@ -5,6 +5,13 @@ replicas:
 image:
   tag: ""
   repository: "nginx"
+resources:
+  limits:
+    cpu: 1
+    memory: 1000Mi
```
alexdesi, line 11: is not the same limit as before?
it'd be great to bumpt it to 2000Mi if possible

alexdesi (approved):
✅ it looks good, just some concern about the 1000Mi (just 1GB)

helm_deploy/values/dev.yaml
sophiebrown01: They'll just get the default which is what is happening now anyway. I think for dev given that we're hardly making requests to it we can probably proceed as we are for now, but we can always revisit this in the future too. I've applied to uat and prod as uat is the closest staging environment we have.

helm_deploy/values/production.yaml
sophiebrown01: It is the same limit as before, `1000Mi`. The difference is that the previous request was only `100Mi`, which is now `500Mi`.

sophiebrown01: As mentioned on Slack too, the limit is not a guarantee of what you'll get, only the request is. If there's enough available capacity in the node and you go over your request threshold, you may be allowed to use up to your limit, but no guarantee. I also don't see why we'd need to be using up to 2Gi per VCD pod, if we're using that much we should look to make improvements.

PR #3263, opened 2026-07-20
ACD-1474: update gem dependencies

alexdesi:
Bumps Ruby gems to their latest compatible versions via `bundle update`


sophiebrown01 (commented):
Looks good. Just one question: what's the reasoning for switching from the RSpec `match` matcher to the `include` matcher?

sophiebrown01 (approved):
Looks good 🥇

PR #3266, opened 2026-08-04
[ACD-1214] Populate the `Needs Linking` tab on the `Link Migrated Cases` page

leandroalemao:
### Ticket

[ACD-1214](https://dsdmoj.atlassian.net/browse/LAACONNECT-1214)

### What

- As a caseworker I need to see all the cases that failed the auto-linking process so that I can manually link them
- Populate the `Needs Linking` tab on the `Link Migrated Cases` page
- The `Link MAAT ID` has been added as per the [Figma design ](https://www.figma.com/design/dO8lRHyMmuzzokgaab9jZO/VCD---DLRM?node-id=4138-2502&m=dev)but as we still don't have the `Link Court Data` page created.... temporarily I linked the `Defendant Link MAAT ID` page instead (which as far as I can see does kind the same thing and link a MAAT ID?). There's a [TODO comment](https://github.com/ministryofjustice/laa-court-data-ui/pull/3266/changes#diff-acab90b7b7eb031ac53ade7edf349226c841c879c3d727071313ef21c5cbe736R31) to change the URL when we get the `Link Court Data` page created.
- The tag in `Case URN` column will be added in the next [PR](https://github.com/ministryofjustice/laa-court-data-ui/pull/3270), as we didn't have the tag added in the Figma page yet by the time this PR was created.

image


spec/views/link_migrated_cases/index.html.haml_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+  let(:cases_count) do
+    { 'action_required' => 3, 'manually_linked' => 5, 'auto_linked' => 2, 'pending' => 7 }
+  end
+
+  let(:cases) do
+    [
+      { 'case_urn' => 'TEST12345', 'defendant_first_name' => 'John', 'defendant_last_name' => 'Smith',
+        'xhibit_case_number' => 'X123', 'court_name' => 'Southwark', 'mode_of_trial' => 'Summary',
+        'maat_id' => '1234567', 'defendant_date_of_birth' => '01/01/1990', 'linked_at' => '2024-03-01',
+        'linked_by' => 'Jane Doe', 'process_errors' => 'Missing case details' }
```
alexdesi: process_errors is actually a hash, not just a plain string. Check out 
CDA branch ACD-1356-migrated-case-process to see the format.

alexdesi: It is stored as jsonb in the table

app/controllers/link_migrated_cases_controller.rb
```diff
@@ -8,10 +8,10 @@ class LinkMigratedCasesController < ApplicationController
                         court_name mode_of_trial reason_for_man_linking maat_id
                         defendant_date_of_birth linked_at linked_by].freeze
   CASES_PER_PAGE = 10
-  TABS = %w[need_linking pending manually_linked auto_linked].freeze
+  TABS = %w[action_required pending manually_linked auto_linked].freeze
```
sophiebrown01, line 11: Curious as to why we're renaming the tab here, unless it's for code readability? The page in the UI is still to be called `Need linking` as per the design

sophiebrown01 (commented):
Looks good! 🙂 

In addition to the one question I posted as a comment, I don't see in this PR that we're linking to the UI page to allow caseworkers to be able to link a particular migrated case to the Common Platform case - on the design this is represented as the _Link MAAT ID_ link for individual cases on the _Need linking_ tab. Just checking if this has been omitted intentionally (because that linking page does not yet exist)?

spec/views/link_migrated_cases/index.html.haml_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    it 'renders the no results message' do
+      render
+      expect(rendered).to include('No link migrated cases found')
+    end
+
+    it 'does not render pagination' do
+      render
+      expect(rendered).to have_no_css('.moj-pagination')
+    end
+  end
```
alexdesi, line 46: Could you move the shared example to the end (for readability)
I am not sure about the message "No link migrated cases found" for all the tags ... but leave it for now, we check it later.

app/controllers/link_migrated_cases_controller.rb
alexdesi: Good question,  he is matching the values of the `status` attribute in the backend: 
```
  enum :status, {
    pending: "pending",
    auto_linked: "auto_linked",
    manually_linked: "manually_linked",
    action_required: "action_required",
  }

```

action_require should be "Need Linking" ...  if I'm not wrong

spec/views/link_migrated_cases/index.html.haml_spec.rb
leandroalemao: Hi @alexdesi I thought I'd be a good idea to move the shared example to a proper `shared_examples` folder. So I've moved it.. please let me know what you think.. tks

app/controllers/link_migrated_cases_controller.rb
leandroalemao: Hi @sophiebrown01 yeah that's the reason @alexdesi said.. in the backend the status of the records we display in the `Need Linking` tab is `action_required`. The tab name itself in the FE is still called 'Need Linking'. I hope it's all good. tks

app/helpers/link_migrated_cases_helper.rb
```diff
@@ -13,20 +13,37 @@ module LinkMigratedCasesHelper
...
-    'linked_by' => { width: '120px', sortable: true, i18n_key: 'linked_by' }
+    'linked_by' => { width: '120px', sortable: true, i18n_key: 'linked_by' },
+    'link_maat_id' => { width: '120px', sortable: false, i18n_key: 'link_maat_id' }
   }.freeze
 
   def column_config(col)
     COLUMN_CONFIG[col]
   end
 
+  def formatted_process_errors(process_errors)
```
pezholio, line 24: No biggie - but it might be good to have some tests for these helper methods. I know there's no tests currently for this helpers, and the behaviour is covered in the view specs, but might be clearer to test the helpers directly too.

pezholio (approved):
Largely looks good to me. One minor point about testing, but not a huge deal/blocker.

app/helpers/link_migrated_cases_helper.rb
leandroalemao: Agree @pezholio I've just added specs for the helper methods.. tksss

app/controllers/link_migrated_cases_controller.rb
sophiebrown01: The question then becomes why is the status in the backend different to the name of the page for `Need linking` when for all of the other pages it matches between the page name and the status? It would be good for consistency to use the same name everywhere but if everyone agrees that it should actually be called `action_required` then I'm also happy to go with this name instead 🤷‍♀️

spec/fixtures/stubs/cda/internal_v2_link_migrated_cases_response.json
```diff
@@ -201,10 +201,10 @@
...
       "linked_at": null,
       "linked_by": null,
-      "process_errors": [
-        "Multiple MAAT applications found",
-        "Manual review required"
-      ]
+      "process_errors": {
+        "error": "429",
+        "message": "Too Many Requests"
+      }
```
sophiebrown01, line 206: Just for reference, this will never be a technical error (such as displaying HTTP response status codes) but one of a set list of responses, based on the auto-linking.

It will either be because no MAAT application was found, multiple MAAT applications were found, the case is already linked (to a Common Platform case) and possibly one other scenario that escapes me too.

app/controllers/link_migrated_cases_controller.rb
leandroalemao: Hi @sophiebrown01 the DLRM Implementation Analysis document states the status needs to be set as `MANUAL_ACTION_REQUIRED`.. maybe that's why the status is called `action_required` in the backend. I reckon it's ok to keep the column as `Need linking` in the FE, we can change it later if that starts to be a problem I guess? tks

sophiebrown01: It definitely needs to be `Need linking` in the frontend as that what's been agreed on the design 🙂 I'm happy to go with the approach here.

PR #3267, opened 2026-08-05
Fix SimpleCov deprecation messages

leandroalemao:
#### What

- Fix SimpleCov deprecation messages

#### Ticket

- N/A

#### Why

- When running specs we can see lots of deprecation messages from SimpleCov.
  One example of deprecation message we can see:
```
[DEPRECATION] `SimpleCov.add_filter` is deprecated. Replace with `SimpleCov.skip` (same arguments, same behavior). Example: `SimpleCov.skip "lib/tasks"`.
```


sophiebrown01 (approved):
Thanks Leandro for helping to clean-up the codebase and keep it current 🚀

PR #3269, opened 2026-08-07
Fix Breadcrumb builder specs

pezholio:
By making the test a `helper` spec, we can pass the `helper` variable into the builder, and the tests run correctly. As we’re doing string matching, we need to include the `role` and `aria-label` in the expectations for the outer `div` test too.


leandroalemao (approved):
👍 👏 tksss

PR #3270, opened 2026-08-11
[ACD-1213] Populate the `Manually linked cases` tab on the `Link Migrated Cases` page

leandroalemao:
### Ticket

[ACD-1213](https://dsdmoj.atlassian.net/browse/LAACONNECT-1213)

### What

- As an NCAT caseworker I want to see which cases have been manually linked;
- Populate the `Manually linked cases` tab on the `Link Migrated Cases` page;
-  Add tag to `Case URN` column in the `Needs Linking` tab;

Manually linked cases tab:
Screenshot 2026-08-11 at 12 00 35

Needs Linking` tab:
Screenshot 2026-08-11 at 12 02 31


app/helpers/link_migrated_cases_helper.rb
```diff
@@ -36,15 +46,18 @@ def link_maat_id_url(defendant_id, case_urn)
...
-                                m_case['defendant_last_name']].compact.join(' ')
-    when 'auto_linked_at' then m_case['linked_at']
-    when 'case_urn_new_tab' then m_case['case_urn']
-    when 'reason_for_man_linking' then formatted_process_errors(m_case['process_errors'])
-    when 'link_maat_id' then link_maat_id_url(m_case['defendant_id'], m_case['case_urn'])
-    else m_case[column]
+    if (handler = HANDLERS[column])
+      send(handler, m_case)
+    else
+      m_case[column]
```
pezholio: I think I'd prefer to stick with a `case` statement for this, rather than using `send` - it makes for a bit too much misdirection. I do think having the behaviour in their own methods is an improvement though.

.rubocop.yml
```diff
@@ -264,3 +264,8 @@ Style/StringLiterals:
 # The new Ruby version 3.4 enforces frozen string literals by default
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+# We should avoid cops that are based on heuristics, since
+# it's not clear what action to take to fix an issue.
+Metrics/ModuleLength:
```
pezholio: Big fan of disabling this cop - it's more annoying than anything!

pezholio (commented):
On the whole I think this looks good - I'm a liiiiitle bit uncomfortable about the metaprogramming, but feel free to push back!

app/helpers/link_migrated_cases_helper.rb
leandroalemao: Hi @pezholio yeah I agree with you.. I prefer the `case` statements as well.. but I was having a battle with `rubocop` so I had to refactor it. I've put it back now how it was before as I disabled a couple of cops from `rubocop`. Also opened this [PR](https://github.com/ministryofjustice/laa-court-data-ui/pull/3272) to actually add govuk rubocop gem which disables lots of these "annoying" cops. tksss

pezholio (approved):
LGTM 👍

app/helpers/link_migrated_cases_helper.rb
sophiebrown01: I agree on the metaprogramming - it is fun to do but we should definitely be putting clarity and understanding first. If Rubocop is complaining about a particular issue and is only offering metaprogramming as an alternative then we can look to ignore those specific cases as they come up.

PR #3271, opened 2026-08-11
Use capybara-playwright for Feature tests

pezholio:
This uses [capybara-playwright](https://github.com/YusukeIwaki/capybara-playwright-driver) as the Capybara driver, which takes advantage of Playwright features like auto-waiting etc. At the moment, `axe-rspec` is incompatible with the Playwright Driver, so I've ~~nicked~~ borrowed a class I used on another project that runs `axe` directly and reads the output from the JS console.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,23 @@ references:
         - node_modules
         - app/assets/builds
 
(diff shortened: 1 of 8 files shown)


leandroalemao (approved):
Bril 👍 👏 tksss

PR #3272, opened 2026-08-11
Add govuk rubocop gem

leandroalemao:
#### What

- Add rubocop govuk gem https://github.com/alphagov/rubocop-govuk

#### Ticket

N/A

#### Why

- rubocop-govuk gem defines the linting rules for GDS Ruby applications, primarily those associated with GOV.UK.


pezholio (approved):
Ah, actually. This is already in the adaptor, so makes sense to use the same approach

PR #3273, opened 2026-08-14
[LAACONNECT-1489] Add link XHIBIT migrated case page

leandroalemao:
#### Ticket

[ACD-1489](https://dsdmoj.atlassian.net/browse/LAACONNECT-1489)

#### Why

- As a caseworker I need to manually link cases that have failed the auto-linking process so that Legal Aid processing can continue

#### What

- Add a new link XHIBIT migrated case page following Figma design;
- It can be accessed clicking on `Link MAAT ID` in any of the items from `Link migrated cases` page `Need linking` tab;
- It depends on this [PR in CDA](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/1592) as it uses a new endpoint added to the backend;

- Updated screenshot:
Screenshot 2026-08-20 at 11 06 07


pezholio (commented):
Not had a proper look at this yet, but looks good from a first glance. Would be good to break it up into a couple of smaller commits though if you can? I think I'd find it easier to see what's going on, especially as I'm so new to the codebase. Ta 👍

app/views/link_migrated_cases/show_link.html.haml
```diff
@@ -0,0 +1,101 @@
...
+          %td.govuk-table__cell
+            = @migrated_case.case_urn
+        %tr.govuk-table__row
+          %th.govuk-table__header{ scope: "row" }
+            = t(".arrest_summons_number")
+          %td.govuk-table__cell
+            = @defendant.arrest_summons_number
+        %tr.govuk-table__row
+          %th.govuk-table__header{ scope: 'row' }
+            = t('.maat_number')
```
sophiebrown01: Going by the design I don't think we need MAAT ID: the case will never be linked on this page so it will always show as not linked

app/views/link_migrated_cases/show_link.html.haml
```diff
@@ -0,0 +1,101 @@
...
+            = t(".reason_for_man_linking")
+          %td.govuk-table__cell
+            = formatted_process_errors(@migrated_case.process_errors)
+
+.govuk-grid-row
+  .govuk-grid-column-full
+    %turbo-frame#offence-table{ src: offences_defendant_path(@migrated_case.defendant_id, urn: @migrated_case.case_urn) }
+      = render partial: 'link_migrated_cases/offences', locals: { defendant: @defendant, offence_history_collection: @offence_history_collection }
+
+- if @defendant.maat_reference.blank?
```
sophiebrown01: Again I think this will always be the case on this linking page as we're only using it for linking (as per the design)

app/views/link_migrated_cases/show_link.html.haml
```diff
@@ -0,0 +1,101 @@
...
+        %tr.govuk-table__row
+          %th.govuk-table__header{ scope: "row" }
+            = t(".reason_for_man_linking")
+          %td.govuk-table__cell
+            = formatted_process_errors(@migrated_case.process_errors)
+
+.govuk-grid-row
+  .govuk-grid-column-full
+    %turbo-frame#offence-table{ src: offences_defendant_path(@migrated_case.defendant_id, urn: @migrated_case.case_urn) }
+      = render partial: 'link_migrated_cases/offences', locals: { defendant: @defendant, offence_history_collection: @offence_history_collection }
```
sophiebrown01: Going by the screenshot on this PR, I think it's missing `Verdict` (as per the design)

app/helpers/link_migrated_cases_helper.rb
```diff
@@ -22,17 +22,16 @@ def column_config(col)
...
     formatted_values.join(' - ').presence || process_errors.to_s
   end
 
-  def link_maat_id_url(defendant_id, case_urn)
-    # TODO: Replace this link when the "Link court data" page is created
-    # Using Defendant Link MAAT ID page instead for now
+  def link_maat_id_url(id)
     link_to("Link MAAT ID",
-            link_defendant_path(defendant_id, urn: case_urn),
+            link_link_migrated_case_path(id),
```
sophiebrown01, line 34: Just checking if this naming here is correct for the path name?

app/helpers/link_migrated_cases_helper.rb
```diff
@@ -22,17 +22,16 @@ def column_config(col)
   end
 
   def formatted_process_errors(process_errors)
-    return process_errors unless process_errors.is_a?(Hash)
+    return process_errors unless process_errors.as_json.is_a?(Hash)
```
sophiebrown01, line 25: Is `process_errors` not already a hash?

app/helpers/link_migrated_cases_helper.rb
```diff
@@ -110,7 +109,7 @@ def handle_reason_for_man_linking(m_case)
   end
 
   def handle_link_maat_id(m_case)
-    link_maat_id_url(m_case['defendant_id'], m_case['case_urn'])
+    link_maat_id_url(m_case['id'])
```
sophiebrown01: Curious as to why we're not using the defendant id and case urn here (also, what `id` is this)?

app/views/link_migrated_cases/_offences.html.haml
```diff
@@ -0,0 +1,36 @@
...
+%table.govuk-table{ class: "govuk-!-margin-bottom-4" }
+  %caption.govuk-table__caption.govuk-visually-hidden
+    = t('search.result.offence.caption')
+  %thead.govuk-table__head
+    %tr.govuk-table__row
+      %th.govuk-table__header{ scope: 'col', class: 'govuk-!-width-one-quarter' }= t('search.result.offence.date')
+      %th.govuk-table__header{ scope: 'col', class: 'govuk-!-width-one-quarter' }= t('search.result.offence.title')
+      %th.govuk-table__header{ scope: 'col', class: 'govuk-!-width-one-quarter' }= t('search.result.offence.plea')
+      %th.govuk-table__header{ scope: 'col', class: 'govuk-!-width-one-quarter' }= t('search.result.offence.mode_of_trial')
+  %tbody.govuk-table__body
```
sophiebrown01, line 16: Ah I think it's in this table we need to add `Verdict` as another column after `Mode of trial`

spec/factories/cda/laa_application.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   # ActiveResource Factory, use :build not :create to prevent HTTP calls
-  factory :laa_application, class: 'Cda::BaseModel' do
+  factory :laa_application, class: 'Cda::LaaApplication' do
```
sophiebrown01: Was this an existing error? Just checking that this wasn't intentional

sophiebrown01 (commented):
Looks good Leandro! A few comments but nothing major 🙂

app/helpers/link_migrated_cases_helper.rb
alexdesi: it should be a `ProcessErrors` object (created by active resource)
but it looks complex,
it seems it could be a hash, a string, nil ... I wonder if we could simplify it 🤔

alexdesi: the process_errors field is a JSONB in cda

alexdesi: I have checked 
the only possible values stored in CDA are:
- nil: never written (the normal state)
- Hash: {"message" => ...} or {"error" => ..., "message" => ...}

app/models/cda/case_summary.rb
```diff
@@ -0,0 +1,4 @@
+module Cda
+  class CaseSummary < BaseModel
+  end
+end
```
alexdesi, line 4: why do you need this? 🤔

app/models/ability.rb
```diff
@@ -68,6 +68,7 @@ def can_search
   def can_manage_links
     can :create, :link_maat_reference
     can :index, :link_migrated_case
+    can %i[show], Cda::LinkMigratedCase
```
alexdesi: For consistency, `can :show,...`?

spec/factories/cda/court_application.rb
```diff
@@ -0,0 +1,19 @@
...
+    application_status { "DRAFT" }
+    application_title { "Appeal against a conviction" }
+    application_type { "String" }
+    received_date { "2025-04-11" }
+
+    case_summary { FactoryBot.build(:case_summary) }
+    hearing_summary { [FactoryBot.build(:hearing_summary)] }
+    subject_summary { FactoryBot.build(:subject_summary) }
+  end
+end
```
alexdesi, line 19: do we need this? 
I wonder why we've never needed to create the factory bot for CourtApplication 🤔

spec/factories/cda/laa_application.rb
leandroalemao: Hi @sophiebrown01 not sure but I feel like we need to link the correct model class in the factory.. not the `Cda::BaseModel`.

app/helpers/link_migrated_cases_helper.rb
leandroalemao: Hi @sophiebrown01 we were reusing the link maat id from defendants page as a placeholder... which requires `defendant_id` and `case_urn` as a param.. now we have a specific link maat id page for migrated cases.. so passing only the `id` will do as we have other information needed in the migrated case record itself.

app/helpers/link_migrated_cases_helper.rb
leandroalemao: Hi @sophiebrown01 CDA sends `process_errors` as a `Hash` but VCD parses as an `Cda::LinkMigratedCase::ProcessErrors` object.. then added `as_json` to parse it back as a Hash.

app/controllers/link_migrated_cases_controller.rb
```diff
@@ -1,8 +1,12 @@
...
 
 class LinkMigratedCasesController < ApplicationController
-  authorize_resource class: false
-  before_action :check_feature_flag
+  authorize_resource class: false, except: %i[show_link link]
+  before_action :load_and_authorize_access,
+                :load_defendant,
+                :load_prosecution_case,
+                :load_offence_histories, only: %i[show_link link offences]
+  before_action :check_feature_flag, :set_breadcrumbs
```
alexdesi: Important: `feature-flag` check now runs after four CDA API calls

But we want to check it first (no need to perform operations if the FF is disabled)

alexdesi (approved):
✅ it looks good to me

app/models/cda/case_summary.rb
leandroalemao: Hi @alexdesi It's used in the factories.. the idea was to improve the test coverage.

spec/factories/cda/court_application.rb
leandroalemao: Hi @alexdesi the idea was to improve the test coverage. 👍 😄

PR #3274, opened 2026-08-20
Fix deprecation messages

leandroalemao:
#### Ticket

N/A

#### What

- We can see some deprecation messages when we start up a server or run specs:

```
DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.

You can emulate the previous behavior with `class_attribute`.
 (called from <top (required)> at /laa-court-data-ui/config/application.rb:19)
DEPRECATION WARNING: `config.active_support.to_time_preserves_timezone` is deprecated and will be removed in Rails 8.2 (called from <top (required)> at /laa-court-data-ui/config/environment.rb:5)
```

#### Why

- `govuk_design_system_formbuilder` was outdated and calling `ActiveSupport::Configurable`. Please see: https://github.com/x-govuk/govuk-form-builder/issues/584

- `active_support.to_time_preserves_timezone` is now set as default to `:zone`. So no need for it anymore. Please see: https://dev.to/lightflight/rails-8-new-activesupport-timezone-defaults-321c

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ gem "govuk_notify_rails", "~> 3.0.0"
 
 gem "govuk-components"
 # rails GDS design system form builder
-gem "govuk_design_system_formbuilder", "~> 5.11"
+gem "govuk_design_system_formbuilder", "~> 6.4"
 gem "haml-rails", "~> 2.1.0"
 gem "lograge"
 gem "oauth2", "~> 2.0.17"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,10 +198,10 @@ GEM
       html-attributes-utils (~> 1.0.0, >= 1.0.0)
       pagy (>= 6, < 10)
       view_component (>= 4.9, < 4.13)
-    govuk_design_system_formbuilder (5.13.0)
-      actionview (>= 6.1)
-      activemodel (>= 6.1)
-      activesupport (>= 6.1)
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -19,7 +19,7 @@ gem "govuk_notify_rails", "~> 3.0.0"
 
 gem "govuk-components"
 # rails GDS design system form builder
-gem "govuk_design_system_formbuilder", "~> 5.11"
+gem "govuk_design_system_formbuilder", "~> 6.4"
```
alexdesi, line 22: thanks for updating this @leandroalemao 
but it's a major release, so I believe there are some incompatibilities, but no other files has been changed.

I found these potential incompatibilities online,
check if we need to make some changes: 

**Changes:** 
```
- Toolchain floors (Bundler-level)
- Injected content is no longer announced (#580, 6.0.0)
- Generated IDs for negative values (#608, 6.0.0)
- Password input markup churn (6.2.1 → 6.4.0)
- GOV.UK Frontend 6 assets
```

leandroalemao: Hi @alexdesi thanks for listing possible incompatibilities... I've updated some `govuk-frontend` packages as well to keep it all inline version wise. I hope it's all good. Thanks.

pezholio (approved):
Had a quick smoke test of the branch locally and all seems in order :+1:

